PR #9, opened 2018-04-27
Some probably non-ideal packaging changes to allow us to build

dsolt:
spindle in one directory while planning to package into an RPM which will
be installed in a different directory.   So we add the configure option
--staging_prefix which maybe is itself poorly named.  The idea is that
it --staging_prefix can be used to set the final install location.   In this
case, many of the directories under this are assumed (especially libexec
and libexec/spindle).  I know there are other configure options to change
the details of where things are placed for --prefix, but our concern is with
things that are compiled into the code that refer to directories and these
are often hard coded with this change.  I would love to see a real solution
to this general problem of how to build spindle when it will later be installed
in a different location from where it is built.  I am not a configure/Makefile
expert obviously.

Example configure:

./configure \
  --enable-sec-none \
  --disable-testsuite \
  --prefix=$installdir \
  --staging_prefix=/opt/ibm/spindle

### DIFF
--- a/configure
+++ b/configure
@@ -6,6 +6,7 @@
 #
 #
 # Copyright (C) 1992-1996, 1998-2012 Free Software Foundation, Inc.
+# Copyright (C) 2018 International Business Machines Corp. 
 #
 #
 # This configure script is free software; the Free Software Foundation
@@ -885,6 +886,7 @@ exec_prefix=NONE
 no_create=
 no_recursion=
 prefix=NONE
+staging_prefix=NONE
 program_prefix=NONE
 program_suffix=NONE
 program_transform_name=s,x,x,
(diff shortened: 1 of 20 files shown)


mplegendre:
Would DESTDIR solve this?  Spindle already supports DESTDIR, and that's generally standard across autotools packages.  You can use it with something like:

% $SRC/configure --prefix=/opt/ibm/spindle ...
...
% make
% make DESTDIR=/some/buildfarm/stagearea install

This immediately produces a spindle install in /some/buildfarm/stagearea/opt/ibm/spindle, which can be used for packaging, but it's configured to eventually run from /opt/ibm/spindle.  


Is this in response to the logging daemons not being found, which you mentioned in email?  I'm not sure what's happening there, but if everything were working the process would be:

1) The configure run produces $BUILD/Makefile, which will contain a line setting pkglibexec to the final install location, based on the -libexecdir or -prefix options passed to configure.
2) The three Makefiles found in $SRC/src/[client, server, fe]/logging/Makefile.am turn that into a define by passing -DLIBEXEC=${pkglibexecdir} to the compiler when building spindle_logc.c
3) $SRC/src/logging/spindle_logc.c:45 turns LIBEXEC into an embedded path in the install, which is later used to fork/exec the logging daemons when SPINDLE_DEBUG is enabled.

I'm not sure where something could be going wrong.  It's odd, because the spindle_bootstrap installation location is tracked in the same way, and if that wasn't found then nothing would work in spindle.

dsolt:
I will see if DESTDIR does what we want and get back to you in a bit.

mcfadden8:
I was able to reproduce the logging daemons not being found problem on my configuration yesterday and have documented it in https://github.com/hpc/Spindle/issues/12

Issue #12, opened 2018-05-10
spindle-build not handling --prefix=path properly for path strings that contain macros which get expanded by CPP


mcfadden8:
This is discussed some in pull request https://github.com/hpc/Spindle/pull/9.

I discovered yesterday that when I used prefix=/foo/bar/linux-rhel7 for configure, the C preprocessor was expanding that to "/foo/bar/1-rhel7".  This is because "linux" was itself a macro which expanded to "1" in the code when $SRC/src/logging/spindle_logc.c was pre-processed.

mplegendre:
Fixed.

PR #13, opened 2018-05-10
Fixed logging build/implementation to properly interpret --prefix=path for log daemon locations

mcfadden8:
Fixed logging build and implementation files to treat the path part of the --prefix=path argument as a string by encapsulating the path in quotation marks.

This addresses issue #12


mcfadden8:
Done.

gpaulsen:
So what's the difference in intent between hpc:devel and hpc:master branches?

PR #19, opened 2018-09-06
Properly NULL-terminate string containing PID of daemon to kill

mcfadden8:
The clang compiler produced a warning on line 131, which upon examination, would cause the pids string to not properly be null-terminated.


mcfadden8:
@mplegendre, please review.

PR #25, opened 2018-11-29
ldcs_hash: return newly added entry to prevent excess lookup

sam6258:
This is a performance improvement to the ldcs cache. If the cache is very large, lookup times can be quite a burden. This removes an excess lookup. In my testing of the cache, this shaves about 2 seconds off of a 14 second init at very small scale.

### DIFF
--- a/src/server/cache/ldcs_hash.c
+++ b/src/server/cache/ldcs_hash.c
@@ -36,7 +36,7 @@ ldcs_hash_key_t ldcs_hash_Val(const char *str) {
    return hash;
 }
 
-void ldcs_hash_addEntry(char *dirname, char *filename) {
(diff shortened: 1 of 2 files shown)


sam6258:
@mplegendre or @jjhursey could you add the appropriate reviewers/tags to this PR? I don't have the correct permissions to do so. Thanks.

mplegendre:
I'll review this when it's ready.  But looks like this is still a work-in-progress, correct?

sam6258:
We will probably have more PR's coming, but this should be good to go for this PR. The purpose of this PR is to just remove the excess hash lookup inside of addEntry.

mplegendre:
I think a commit is missing here.  This just changes the return type of the hash function, but it doesn't remove any excess hash lookups.

PR #31, opened 2020-05-12
If we are preloading files, also preload 'stat' result

jjhursey:
* Enabled by default, can be disabled by setting the envar:
   `SPINDLE_DISABLE_PRELOAD_STAT=1`


jjhursey:
When preloading files, we noticed that there is a `stat` call before the file is opened (to check permissions). This requires 2 calls to the root server to (1) cache the `stat` result, then later (2) to cache the file. By preloading the `stat` result with the file we combine those calls thus reducing the time to preload files.

src/server/auditserver/ldcs_audit_server_handlers.c
```diff
@@ -1609,8 +1613,17 @@ static int handle_preload_filelist(ldcs_process_data_t *procdata, ldcs_message_t
...
+
+   if( is_stat_enabled < 0 ) {
+       const char *envar = getenv("SPINDLE_DISABLE_PRELOAD_STAT");
+       is_stat_enabled = 1; // Default
+       if( NULL != envar && '1' == envar[0] ) {
+           is_stat_enabled = 0;
+       }
+   }
+
    debug_printf2("At top of handle_preload_filelist\n");
```
mplegendre, line 1627: What's the motivation behind making this disable-able via environment variable?  People won't get this if they don't preload, so it can already be avoided if someone's grossly offended by stats.

jjhursey: It was mostly to have a way to disable it if it became a problem for those enabling preloading. That being said I don't think we ever had to use it to disable the feature.

It can be used to easily see the benefit of the optimization. But that's more of a developer consideration than a user consideration.

I'm ok if you would prefer to remove it.

mplegendre:
Closing old PRs

PR #33, opened 2020-05-12
Remove `lib` prefix check from spindle stat

jjhursey:
* With this in place `mkdir` and `stat` on directories and non-DSO files
   will fail. Encountered when working with a `libcoll_cache` directory.

### DIFF
--- a/src/client/client/should_intercept.c
+++ b/src/client/client/should_intercept.c
@@ -187,8 +187,7 @@ int stat_filter(const char *fname)
    last_slash = strrchr(fname, '/');
 
    if (is_dso(fname, last_slash, last_dot) ||
-       is_python(fname, last_dot) || 
-       is_lib_prefix(fname, last_slash))
+       is_python(fname, last_dot))
       return REDIRECT;
    else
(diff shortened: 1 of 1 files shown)


jjhursey:
Thanks @sam6258 for finding and fixing this issue.

jjhursey:
An alternative to this would be to adjust `is_lib_prefix` to:
 * Check for `lib` prefix, and
 * Check that the `fname` is a file, and
 * Check that the `fname` is suffixed with either `.so` or `.a`
I don't know if that is too heavyweight for this check or not.

mplegendre:
This got fixed in the last release by adding the .so suffix test--I also hit the libcoll_cache issue.  On more consideration, we could be even safer by checking on file read that the lib*.so* is really a DSO, and canceling the broadcast if not.  I'll open a new issue for that.

mplegendre:
https://github.com/hpc/Spindle/commit/22871945557038ba375c8f1eea1f20d090ee723d -- the fix

jjhursey:
Perfect. Thanks. I didn't notice that.

PR #34, opened 2020-05-12
Exclude local path prefixes from being cached

jjhursey:
* Feature enabled by default, but can be disabled by setting the
   envar `SPINDLE_DISABLE_EXCLUDE=1`

### DIFF
--- a/src/client/client/client.c
+++ b/src/client/client/client.c
@@ -37,6 +37,7 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include "client_api.h"
 #include "spindle_launch.h"
 #include "shmcache.h"
+#include "should_intercept.h"
 
 errno_location_t app_errno_location;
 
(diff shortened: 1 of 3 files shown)


jjhursey:
This one I would like to get some feedback on. I don't love the explicit list of file system mounts. Something better might be to check the file system associated with the path to exclude the local file system mounts. I wasn't sure how to do that portably / efficiently.
I'm mostly posting this for comment/discussion.

mplegendre:
I like the feature, but agree about needing a better implementation.  Some of those hardcoded paths are NFS on our linux clusters.  

The last release added an internal capability to parse the system's mount points in the server: 
https://github.com/hpc/Spindle/blob/devel/src/server/auditserver/parse_mounts.cc

We're currently using this to correct the device field in stat calls when different nodes have different device major/minor ids.   But we could also parse a prefix list out of this that identifies local filesystems.  Then have servers respond to clients with a message to self-read if it sees something prefixed as local.

We'd need a proper list of what filesystem types are remote vs local.

mplegendre:
Closing old PRs. This got implemented as a configurable feature some time ago.

PR #52, opened 2024-02-19
Enhance and fix SPANK plugin

neicker:
Besides fixing several bugs in the context of the SPANK plugin this PR mainly introduces the utilization of `spank_prepend_task_argv()`.

`spank_prepend_task_argv()` was introduced in Slurm 23.11 and allows to manipulate the (to be spawned) task's `argv` from within a SPANK plugin. This makes Spindle's SPANK plugin working again with Slurm 23.11 and above.

It seams to me that the SPANK plugin broke once the tweaking of libc's jump table in `spindleHookSpindleArgsIntoExecBE()` and `interceptExecForMap()` stopped working due to making this table read-only.


nchaimov:
These commits were merged as part of #125.

PR #56, opened 2024-11-04
Fixes logging suppression caused by early logging.

### LINKED ISSUE #55: All front-end debugging messages suppressed by early initial message
Calls to `spindle_debug_printf_impl()` (usually via a wrapper) will trigger logger initialization via `init_spindle_debugging()` as needed.  Correct initialization requires the `SPINDLE_DEBUG` environment variable to be already set.  However, this does not happen until several lines into `sp_init()` (when running under Flux).

When that environment variable is not yet set, the initialization process assumes the absence of the variable is intentional and short-circuits further attempts at initialization.

A possible two-part fix:

Force `sp_init()` to always set SPINDLE_DEBUG in the shell environment, setting it to zero if the variable was not present in the flux shell environment.  Force `init_spindle_debugging()` to check for the existence of `SPINDLE_DEBUG`.  If the variable is not present, take that as an indication that shell initialization isn't yet complete.  The message should probably be logged regardless of log level; alternatively, it could be buffered and logged later, but that gets complicated quickly.  Subsequent calls to `spindle_debug_printf_impl()` can check if the environment variable has been set yet, and reset initialization if needed.

rountree:
Logging initialization assumes a missing SPINDLE_DEBUG environment variable means SPINDLE_DEBUG=0.  Messages that are logged prior to that environment variable being set are lost and inadvertently suppress any further logging messages from that subsystem.

This patch sets SPINDLE_DEBUG=0 in the shell environment if that variable is missing from the flux environment.  It also modifies the logging initialization process so that any messages logged before SPINDLE_DEBUG is set are sent through, but any subsequent message will trigger re-initialization until SPINDLE_DEBUG exists.

Fixes #55

### DIFF
--- a/src/flux/flux-spindle.c
+++ b/src/flux/flux-spindle.c
@@ -345,9 +345,11 @@ static int sp_init (flux_plugin_t *p,
     /*  If SPINDLE_DEBUG is set in the environment of the job, propagate
      *  it into the shell so we get spindle debugging for this session.
      */
-    if ((debug = flux_shell_getenv (shell, "SPINDLE_DEBUG")))
+    if ((debug = flux_shell_getenv (shell, "SPINDLE_DEBUG"))){
         setenv ("SPINDLE_DEBUG", debug, 1);
-
+    }else{
+        setenv ("SPINDLE_DEBUG", "0", 1);
+    }
     /*  The spindle testsuite requires SPINDLE_TEST
      */
     if ((test = flux_shell_getenv (shell, "SPINDLE_TEST")))
--- a/src/logging/spindle_logc.c
+++ b/src/logging/spindle_logc.c
@@ -273,24 +273,34 @@ void reset_spindle_debugging()
    init_spindle_debugging(spindle_debug_name, 0);
 }
 
-void init_spindle_debugging(char *name, int survive_exec)
+int init_spindle_debugging(char *name, int survive_exec)
 {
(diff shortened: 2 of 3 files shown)


mplegendre:
We need to leave debugging off if SPINDLE_DEBUG is unset. There's significant overhead/poor-scaling if debugging is enable. And file pollution, as it drops files in your PWD. 
The typical way users will run will be without SPINDLE_DEBUG set. There's also several run modes that don't go through the flux plugin, so we can't just rely on setting/unsetting it there.

If you're going to insert a compile-time instrumentation that requires debug tracing, then add a compile-time flag that turns debugging on by default. I can help add that flag, if that would be easier.

rountree:
Point taken wrt scaling and file pollution.  I can imagine a solution where messages (with their priority) are buffered until initialization occurs, then parsed to see if they're logged or discarded.  It's not wholly unlike trying to make `printf()` work in `_start()` before your file descriptors are set up, and, perhaps like that, it's not a problem worth solving (or not worth solving this week).

mplegendre:
But then you've got to decide how long to buffer. And you're still paying the internal overhead of logging.

I've got a upcoming commit that changes Spindle's config management. It's already got some other changes to autoconf configure.  I'll throw in a configure --with-default-debugging-level=X with that commit, and we can set that when adding the -finstrument. 

I'm going to close this. We can re-open if the discussion circles back to this kind of approach.

PR #58, opened 2025-01-17
Use --external-launcher for Slurm launch

mattaezell:
The Slurm `srun --external-launcher` option, added in 23.11, allows users to start a daemon that has access to all the resources without actually consuming any of them. It was intended for process managers like hydra, orted, or flux-as-a-step-manager to launch tasks, but I think we can use it here also.

The configury need more work, but I wanted to get some feedback before spending time on that. Nothing older than 23.11 is considered "supported" anymore. Does it make sense to completely remove the `BROKEN_SRUN` construct at this point?

```
ezy@borg005:~/mpi_hello> time spindle srun -n2240 -N40 ./mpihi >/dev/null

real    0m6.724s
user    0m0.021s
sys     0m0.054s
ezy@borg005:~/mpi_hello> time srun -n2240 -N40 ./mpihi >/dev/null

real    0m11.810s
user    0m0.015s
sys     0m0.044s
```

### DIFF
--- a/configure.common.ac
+++ b/configure.common.ac
@@ -123,8 +123,8 @@ AS_CASE([$sysconfdir],
 BROKEN_SRUN=0
 bad_srun_major=20
 bad_srun_minor=11
-good_srun_major=0
-good_srun_minor=0
+good_srun_major=23
+good_srun_minor=11
 if test "x$ENABLE_SLURM" == "xtrue"; then
   AC_PATH_PROG(SRUN_PATH, "srun", "none")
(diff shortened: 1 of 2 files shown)


mplegendre:
I hadn't been aware of the --external-launcher before, but a quick scan of the man page is promising. That is the kind of thing we're trying to do when we launch spindle_be. Will investigate further.

nchaimov:
The commit in this PR was rebased and merged as part of #178.

PR #80, opened 2025-04-21
Fixes hang introduced by logging in sp_init().

### LINKED ISSUE #79: Logging early in `sp_init()` causes hang.
This information log message in `flux-spindle.c` hangs.

https://github.com/hpc/Spindle/blob/d2dd15cd814f1b26b88ca858eb1c65e3e2f747c6/src/flux/flux-spindle.c#L479

The problem is that the log message occurs before a call to `setenv()` (transferring the value of `SPINDLE_TEST` from the flux shell to the regular shell).  The log message triggers initialization of the logging system, which checks for the `SPINDLE_TEST` environment variable, which isn't present, thus `run_tests=0`.  After initialization completes and the message is logged, the remaining code in `sp_init()` executes, which includes setting `SPINDLE_TEST=1`.  Subsequent processes see this different value and set up the test environment, but as the first process never did, the result is a hang.

In the two runs I looked at, only the smallest pid had `run_test=0`, which probably means flux calls `sp_init()` serially.

A simple solution is to move all of the `setenv()` code into `flux_plugin_init()` along with a note to not log anything until all calls to `setenv()` complete.

rountree:
Logging initialization relies on environment variables, and some of those variables originate in the flux shell.  Initiali- zation prior to those values being transferred can result in a hang.

This PR moves the transfer code into flux_plugin_init(), which is the first spindle-provided callback to be executed.

Passes all runTests on Tuolumne.

Fixes #79

### DIFF
--- a/src/flux/flux-spindle.c
+++ b/src/flux/flux-spindle.c
@@ -458,10 +458,6 @@ static int sp_init (flux_plugin_t *p,
     int shell_rank;
     flux_future_t *f;
     json_t *R;
-    const char *debug;
-    const char *tmpdir;
-    const char *test;
-    const char *spindle_enabled;
 
     if (!(shell = flux_plugin_get_shell (p))
         || !(h = flux_shell_get_flux (shell)))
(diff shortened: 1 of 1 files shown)


mplegendre:
Thanks for the reminder on this.

This is already fixed in devel. I moved the SPINDLE_TEST environment handling to above the first debug_print in sp_init() in a recent-ish commit. I think that has the same effect as these changes. 

Moving to flux_plugin_init, as this PR does, is probably better. It'll harder to make a future mistake and put a debug_print to early in sp_init(). But this no longer fixes anything and probably doesn't have the priority it used to. You can either cleanup the conflict and re-submit this--I'll merge it. Or we can just close it out.

PR #95, opened 2025-07-31
Fix warnings

### LINKED ISSUE #94: Clean up warnings

rountree:
Fixes warnings generated by gcc 15.1.0 and `-Wall -Wextra -Werror`.

The vast majority are either unused parameters or comparing incompatible types (signed vs unsigned).  I handled the former by adding `foo=foo;` (or, in C++ code, removing the parameter name from the function signature).  The latter could be a bit more involved, but generally if a variable contains a count of things in memory, it's now `size_t`, with a few workarounds to handle glibc calls that were defined in less suspicious times.

A single case of intentional fallthrough is now marked with `[[ fallthrough ]]`. This feature has been available in gcc since v10.0 and is now part of the C23 standard.  If we have users with older compilers, there are other possible workarounds.

The single (fixed) bug uncovered was an unintentional fallthrough.  

I can chop this up into smaller (per file?) chunks if that's useful.

Fixes #94

### DIFF
--- a/src/biter/client_queue.c
+++ b/src/biter/client_queue.c
@@ -91,6 +91,7 @@ int init_message(int num_procs, void *header_ptr, void *session)
 
 void get_message(int for_proc, void **msg_data, size_t *msg_size, size_t *bytes_read, void *session)
 {
+   session = session;
    message_t *msg;
    sheep_ptr_t *msgp = proc_messages + for_proc;
    assert(!IS_SHEEP_NULL(msgp));
@@ -103,6 +104,7 @@ void get_message(int for_proc, void **msg_data, size_t *msg_size, size_t *bytes_
 
 int has_message(int for_proc, void *session)
 {
+   session=session;
    sheep_ptr_t *msgp = proc_messages + for_proc;
    return !IS_SHEEP_NULL(msgp);
 }
@@ -209,6 +211,7 @@ int get_message_space(size_t msg_size, unsigned char **msg_space, void **header_
 
 void update_bytes_read(int for_proc, size_t newval, void *session)
 {
+   session=session;
    message_t *msg;
    sheep_ptr_t *msgp = proc_messages + for_proc;
    assert(!IS_SHEEP_NULL(msgp));
--- a/src/biter/shmutil.h
+++ b/src/biter/shmutil.h
@@ -17,18 +17,19 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #if !defined(SHMUTIL_H_)
 #define SHMUTIL_H_
 
+#include <sys/types.h>
 #include "sheep.h"
 #include "spindle_launch.h"
 
 typedef struct {
-   volatile unsigned long *lock;
-   volatile unsigned long *held_by;
+   volatile pid_t *lock;
+   volatile pid_t *held_by;
    int ref_count;
 } lock_t;
 
 typedef struct {
    volatile int cur_id;
-   unsigned long locks[2];
+   pid_t locks[2];
 } base_header_t;
 
 #include "demultiplex.h"
@@ -43,11 +44,11 @@ typedef struct {
    int num_ranks;
    int read_file;
    int num_started;
-   unsigned long locks[6];
+   pid_t locks[6];
 } biter_header_t;
 
 typedef struct {
-   unsigned long locks[2];
+   pid_t locks[2];
    sheep_ptr_t hash;
    sheep_ptr_t lru_head;
    sheep_ptr_t lru_end;
--- a/src/client/client/realpath.c
+++ b/src/client/client/realpath.c
@@ -347,7 +347,7 @@ realpath_stk (const char *name, char *resolved, struct realpath_bufs *bufs)
 
       /* Length of this file name component; it can be zero if a file
          name ends in '/'.  */
-      long startlen = end - start;
+      size_t startlen = end - start;
 
       if (startlen == 0)
          break;
@@ -365,7 +365,7 @@ realpath_stk (const char *name, char *resolved, struct realpath_bufs *bufs)
          if (!ISSLASH (dest[-1]))
             *dest++ = '/';
 
-         while (rname + bufs->rname.length - dest
+         while ( (size_t)(rname + bufs->rname.length - dest)
                 < startlen + sizeof dir_suffix)
          {
             long dest_offset = dest - rname;
--- a/src/client/client_comlib/client_api.c
+++ b/src/client/client_comlib/client_api.c
@@ -57,7 +57,7 @@ int send_file_query(int fd, char* path, int dso, char** newpath, int *errcode) {
 
    COMM_LOCK;
 
-   debug_printf3("sending message of type: file_query len=%d data='%s' ...(%s)\n",
+   debug_printf3("sending message of type: file_query len=%zu data='%s' ...(%s)\n",
                  message.header.len, message.data, path);  
    client_send_msg(fd, &message);
 
@@ -105,7 +105,7 @@ int send_stat_request(int fd, char *path, int is_lstat, char *newpath)
    
    COMM_LOCK;
 
-   debug_printf3("sending message of type: %sstat_query len=%d data='%s' ...(%s)\n",
+   debug_printf3("sending message of type: %sstat_query len=%zu data='%s' ...(%s)\n",
                  is_lstat ? "l" : "", message.header.len, message.data, path);  
    client_send_msg(fd, &message);
 
@@ -149,7 +149,7 @@ int send_existance_test(int fd, char *path, int *exists)
    message.header.len = strlen(path) + 1;
    message.data = (void *) buffer;
 
-   debug_printf3("Sending message of type: file_exist_query len=%d, data=%s\n",
+   debug_printf3("Sending message of type: file_exist_query len=%zu, data=%s\n",
                  message.header.len, path);
    COMM_LOCK;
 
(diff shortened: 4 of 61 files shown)


src/client/client_comlib/client_api.c
```diff
@@ -221,17 +221,25 @@ int send_dirlists_request(int fd, char **local_result, char **exece_result, char
    client_recv_msg_dynamic(fd, &message, LDCS_READ_BLOCK);
    COMM_UNLOCK;
 
+   debug_printf("QQQ After client_recv_msg_dynamic, message.header.len=%zu.\n", message.header.len);
```
mplegendre, line 224: Clean up debug printfs through here. If these are generally useful, then format as normal and make them debug_printf3. If not generally useful, then delete.

src/biter/shmutil.h
```diff
@@ -17,18 +17,19 @@ Place, Suite 330, Boston, MA 02111-1307 USA
...
 #define SHMUTIL_H_
 
+#include <sys/types.h>
 #include "sheep.h"
 #include "spindle_launch.h"
 
 typedef struct {
-   volatile unsigned long *lock;
-   volatile unsigned long *held_by;
+   volatile pid_t *lock;
```
mplegendre, line 25: Leave 'lock' an unsigned long (or a uint64_t).  Same with the other 'lock' variables below.

held_by can remain a pid_t

src/client/client/realpath.c
```diff
@@ -347,7 +347,7 @@ realpath_stk (const char *name, char *resolved, struct realpath_bufs *bufs)
 
       /* Length of this file name component; it can be zero if a file
          name ends in '/'.  */
-      long startlen = end - start;
+      size_t startlen = end - start;
```
mplegendre, line 350: This file is copied from the libc implementation of realpath (we're both LGPL, so that's okay). Would rather leave the body of libc's realpath as unmodified as possible.

src/client/client/realpath.c
```diff
@@ -365,7 +365,7 @@ realpath_stk (const char *name, char *resolved, struct realpath_bufs *bufs)
          if (!ISSLASH (dest[-1]))
             *dest++ = '/';
 
-         while (rname + bufs->rname.length - dest
+         while ( (size_t)(rname + bufs->rname.length - dest)
```
mplegendre, line 368: Same as above for not modifying realpath if not necessary

src/cobo/handshake.c
```diff
@@ -147,9 +147,9 @@ static int handshake_wrapper(int sockfd, handshake_protocol_t *hdata, uint64_t s
                              int is_server);
 static int handshake_main(int sockfd, handshake_protocol_t *hdata, uint64_t session_id,
                           int is_server);
-static int reliable_write(int fd, const void *buf, size_t size);
-static int reliable_read(int fd, void *buf, size_t size);
-static int read_key(char *key_filepath, int key_length_bytes);
+static size_t reliable_write(int fd, const void *buf, size_t size, int *err);
+static size_t reliable_read(int fd, void *buf, size_t size, int *err);
```
mplegendre, line 151: Not sure that the 'int *err' is the way to go here. The INTERNAL error that this returns is the error code for something going wrong in the spindle code, implying a spindle bug. But we don't really know the level of error in this kind of low level code. There's non-spindle-bug situations where reliable_read and reliable_write fail.

We could do better error returns by going to a ssize_t  return and returning -1 from these. Then let the higher level code decide what kind of error it is.

src/include/ldcs_api_pipe.h
```diff
@@ -35,8 +35,8 @@ char *ldcs_get_connection_string_pipe(int fd);
 int ldcs_register_connection_pipe(char *connection_str);
 
 /* internal */
-int _ldcs_write_pipe(int fd, const void *data, int bytes );
-int _ldcs_read_pipe(int fd, void *data, int bytes, ldcs_read_block_t block );
+size_t _ldcs_write_pipe(int fd, const void *data, size_t bytes, int *err );
+size_t _ldcs_read_pipe(int fd, void *data, size_t bytes, ldcs_read_block_t block, int *err );
```
mplegendre, line 39: Similar concerns here, though there's larger issues with adding err to ldcs_read_pipe and ldcs_write_pipe. The implementations below set err even when there's soft errors that the functions recover from. But the calling code treats all contents of *err as a hard error. 
Again, suggest making the return ssize_t and using a -1 return on error.

src/server/comlib/ldcs_api_pipe.c
```diff
@@ -342,63 +349,60 @@ int ldcs_recv_msg_static_pipe(int fd, ldcs_message_t *msg, ldcs_read_block_t blo
...
-  dataptr   = (char*) data;
 
   while (left > 0)  {
     btoread    = left;
     debug_printf3("before read from fifo %d, bytes_to_read = %ld\n", fd, btoread);
-    bread      = read(fd, dataptr, btoread);
-    if(bread<0) {
+    errno=0;
+    rc = read(fd, dataptr, btoread);
+    *err = errno;
```
mplegendre, line 391: As an example from the above review comment, here we set *err even if it's not a hard error.

src/server/auditserver/ldcs_audit_server_handlers.c
```diff
@@ -1440,6 +1442,7 @@ static int handle_request_file(ldcs_process_data_t *procdata, node_peer_t from,
       case REQ_DIRECTORY:
          dir_result = handle_send_query(procdata, dirname, 1, 0);
          add_requestor(procdata->file_requests, dirname, from);
+         [[ fallthrough ]]; /* Requires gcc >= 10.0 or -std=c23 or ... */
```
mplegendre, line 1445: Can we ifdef this to the right gcc versions?

src/server/auditserver/ldcs_elf_read.c
```diff
@@ -32,10 +32,9 @@ static int readUpTo(FILE *f, unsigned char *buffer, size_t *cur_pos, size_t new_
       return 0;
 
    do {
+      errno = 0;
       result = fread(buffer + *cur_pos, 1, new_size - *cur_pos, f);
-   } while (result == -1 && errno == EINTR);
-   if (result == -1)
-      return -1;
+   } while (errno == EINTR);
```
mplegendre, line 37: Needs the (result == -1) part of the check. errno only has meaning if the function returns -1

mplegendre (changes requested):

src/biter/client_queue.c
```diff
@@ -91,6 +91,7 @@ int init_message(int num_procs, void *header_ptr, void *session)
 
 void get_message(int for_proc, void **msg_data, size_t *msg_size, size_t *bytes_read, void *session)
 {
+   session = session;
```
SamHerts, line 94: While `foo=foo;` is a valid fix to the unused parameter warning, the standard way of handling this would be to cast void: 
`(void)session;`
Compiler will no-op either way, but the latter is more common and readable.

src/client/client_comlib/client_api.c
rountree: No stray QQQs in the gcc-15.1.0 PR.

src/client/client/realpath.c
rountree: In the gcc-15.1.0 PR I suppressed those warnings in the modified libc code.

src/cobo/handshake.c
rountree: New PR uses `ssize_t` where appropriate.

src/include/ldcs_api_pipe.h
rountree: Might want to take a look at this one.  Here's the code in the new PR (I haven't changed the logic):
```
size_t _ldcs_write_pipe(int fd, const void *data, int bytes ) { 
  size_t      bsumwrote;
  ssize_t     bwrite, bwrote, left;
  char       *dataptr;
  
  left      = bytes;
  bsumwrote = 0;
  dataptr   = (char*) data;

  while (left > 0) {
    bwrite     = left;
    bwrote     = write(fd, dataptr, bwrite);
    left      -= bwrote;
    dataptr   += bwrote;
    bsumwrote += bwrote;
  }
  return (bsumwrote);
}
```
If `write()` returns `-1`, the `left` variable increases and we won't exit until `left` rolls over.

The `_read` has a similar issue, in that errors will return 0 rather that percolate the error value up.

```
size_t _ldcs_read_pipe(int fd, void *data, int bytes, ldcs_read_block_t block ) {
...
    bread      = read(fd, dataptr, btoread);
    if(bread<0) {
       if( (errno==EAGAIN) || (errno==EWOULDBLOCK) ) {
          if (print_count-- > 0)
             debug_printf3("read from fifo: got EAGAIN or EWOULDBLOCK\n");
          if(block==LDCS_READ_NO_BLOCK)
             return 0;
          else
             continue;
       } else { 
          debug_printf3("read from fifo: %ld bytes ... errno=%d (%s)\n",bread,errno,strerror(errno));
       }
    } else {
       debug_printf3("read from fifo: %ld bytes ...\n",bread);
    }
...
```
The `-Wsign-compare` warnings are taken care of, but we might want to revisit the underlying code.

src/server/auditserver/ldcs_audit_server_handlers.c
rountree: The new PR uses a magic comment (`/* Fall through */`) that was introduced when the warning was introduced.  Earlier versions of gcc won't notice the issue, later ones know to look for the comments.

src/server/auditserver/ldcs_elf_read.c
rountree: Weirdly enough, `fread()` doesn't return -1 on error, and doesn't set `errno`.  That might not be what you want.

src/biter/client_queue.c
rountree: @SamHerts Thanks.  I've created a new PR that uses the cast-to-void.

rountree:
Will close this PR as soon as the rest of the discussions are resolved.  The new PR is gcc-15.1.0.

PR #97, opened 2025-08-29
Fix segfault when binding Spindle API functions and enable Spindle API tests

nchaimov:
Fixes a bug where attempts to bind any symbol in the Spindle API would cause a segfault. 

The function `client_call_binding` would attempt to check if `*binding->libc_func` was NULL, and, if so, assign the address of the symbol to it:
https://github.com/hpc/Spindle/blob/d8449f144252170eef77d7596b1e2e134d3bea0b/src/client/auditclient/redirect.c#L39-L40
However, for Spindle API calls, these pointers were NULL (second field below), so dereferencing `binding->libc_func` would crash:
https://github.com/hpc/Spindle/blob/d8449f144252170eef77d7596b1e2e134d3bea0b/src/client/client/intercept.c#L55-L63

The issue is fixed by adding variables to receive the address, as is done for libc functions. 

Note that the NULL is left in place for `spindle_test_log_msg` because that function has special handling in `client_call_binding` which occurs prior to the access of `*binding->libc_func`:
https://github.com/hpc/Spindle/blob/d8449f144252170eef77d7596b1e2e134d3bea0b/src/client/auditclient/redirect.c#L28-L29

There are existing tests for the Spindle API, but `runTests` did not execute them. The `--spindleapi` tests are added to `runTests`.

There was an additional bug where the test driver would check for the results of loading shared libraries, even though these libraries are not loaded with `--spindleapi`. The `check_libraries()` function had an existing check for this:
https://github.com/hpc/Spindle/blob/d8449f144252170eef77d7596b1e2e134d3bea0b/testsuite/test_driver.c#L1058-L1062
This check was also added to `checkTlsSum()`.


mplegendre:
One side effect of this is that spindle is trying to lookup/bind the spindle API functions in libc. That obviously won't ever trigger, but it tries with every library load and is a small overhead. 

Separate from this PR, we may want to put some kind of marking in the binding table to indicate that spindle should not look for these symbols in libc. Spindle also does this with the unused aliases for stat/lstat/fstat. A more general fix might be to mark these as "not in libc" after the first lookup fails.

But things look good within the scope of this PR.

PR #98, opened 2025-09-05
Locations

rountree:
Work in progress:

Broad strokes:

- Removes the "location" parameter
- Creates a "commpath" parameter that takes multiple, comma-separated paths.  Clients use the first valid path in the list to communicate to their server.
- Creates a "cachepath" parameter that takes multiple, comma-separated paths.  Servers arrive at a consensus of what paths are valid across all servers, choosing the first common valid path for the file cache.
- Adds cobo_allreduce and path selection functionality.


rountree:
This suffers from heroic programmer syndrome.  I'm disassembling it into its constituent parts and putting it back together more deliberately.

PR #99, opened 2025-09-09
Cachepath work

rountree:
Providing a `cachepaths=` line in the `~/.spindle/spindle_conf` file with a colon-separated list of directories results in Spindle selecting the first writeable directory in the list common to all servers as the cache directory.

Passes all tests in runTests.

Returning to lessons I learned 30 years ago:  this PR is the minimum required for a working `cachepath` option.  This PR does NOT:

1. Provide a separate, multi-path `commpath` option.  The usual `location` is used for the communication path.
2. Upgrade the `location` parameter to be multi-path.
3. Replace all uses of `location` with `commpath` or `cachepath` as appropriate.
4. Remove all messages/environment variables/arguments associated with `location`.

Any one of those changes was trivially easy to make.  Debugging after having made all of those changes was well beyond what I'm able to do.  Reducing the feature set to just `cachepath` allowed the debugging to be completed in a handful of hours.

I'll be adding the following to this PR:

1. Ability to specify `cachepath` via flux.
2. Ability to specify `cachepath`at configuration time.


mplegendre:
Can you rebase this into either:
A) one commit
B) or better, a small handful of commits that break up the work into distinct component pieces
?

PR #102, opened 2025-09-19
Continuous integration with GitHub Actions

nchaimov:
Adds continuous integration with GitHub Actions.

The current implementation defines two containers, one for building Spindle with `--with-rm=serial` and one for `--with-rm=flux`. The Flux container uses Docker Compose to set up a cluster of 4 nodes. The containers are set up such that when building the image, the Spindle repo is copied into the container and built there. The GitHub Actions workflow checks out the commit to be tested and builds the images incorporating the checked-out commit.

The tests will fail until #97 and #100 are merged, as these fix bugs affecting the Ubuntu environment in the containers. See https://github.com/ParaToolsInc/spindle-testing/actions/runs/17871033110 for an example in a fork which already has these fixes applied to see the tests passing. And see https://github.com/ParaToolsInc/Spindle/actions/runs/17871526378 for what happens with the commits without these fixes, showing that failing tests indeed fail the CI run as expected.

If we want to require manual authorization before running the tests, we will have to first create an environment in GitHub and set the workflow to run in that environment.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+ version: 2 
+ updates: 
+   - package-ecosystem: "github-actions" 
+     directory: "/.github/workflows" 
+     schedule: 
+       interval: "monthly" 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,106 @@
+name: ci
+
+on:
+  push:
+    branches:
+      - devel
+  pull_request:
+    branches:
+      - devel
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ci-${{github.ref}}-${{github.event.pull_request.number || github.run_number}}
+  cancel-in-progress: true
+
+jobs:
+  spindle-serial-ubuntu:
+    name: Testsuite (Serial, Ubuntu)
+    environment: Spindle CI
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Check out Spindle
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
+
+      - name: Setup Docker Compose
+        uses: docker/setup-compose-action@364cc21a5de5b1ee4a7f5f9d3fa374ce0ccde746
+        with:
+          version: latest
+
+      - name: Build spindle-serial-ubuntu image
+        id: serial-ubuntu-build
+        run: |
+          cd containers/spindle-serial-ubuntu
+          docker compose --progress=plain build
+
+      - name: Bring spindle-serial-ubuntu up
+        id: serial-ubuntu-up
+        run: |
+          cd containers/spindle-serial-ubuntu
+          docker compose up -d
+
+      - name: Verify munge works in spindle-serial-ubuntu
+        id: serial-ubuntu-munge
+        run: |
+          docker exec spindlenode bash -c 'munge -n | unmunge'
+
+      - name: Run spindle-serial-ubuntu testsuite
+        id: serial-ubuntu-testsuite
+        run: |
+          docker exec spindlenode bash -c 'cd Spindle-build/testsuite && ./runTests'
+
+      - name: Bring spindle-serial-ubuntu down
+        id: serial-ubuntu-down
+        if: ${{ always() }}
+        continue-on-error: true
+        run: |
+          cd containers/spindle-serial-ubuntu
+          docker compose down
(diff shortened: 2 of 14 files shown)


nchaimov:
With #97 and #100 merged, I've rebased off devel. The tests now pass, as can be seen in the CI run on my fork: https://github.com/ParaToolsInc/Spindle/actions/runs/17885768345

nchaimov:
The CI workflow is now configured to run in the `Spindle CI` environment, which will require authorization before running the CI workflow using the organization's resources. Forks can create an environment with the same name to run using the fork owner's resources. Additionally, I have verified that the CI workflow can be run locally by using [gh-act](https://github.com/nektos/gh-act).

I think this should be ready to merge. The only possibly remaining issue is when to run the CI workflow. As currently configured, it will run on all pull requests, and any commits to any branch in this repo. That will generate a request for review on every commit, but if we don't want that, we could instead run only on pull requests, or only on commits to specific branches, such as to `devel`.

mplegendre:
I'm adding @alecbcs as a reviewer. He's a developer at here and an expert on how we manage CI. 
Alec, help/guidance would be appreciated here. And we can talk more about this tomorrow if you have questions.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,97 @@
+name: CI
+
+on: [push, pull_request, workflow_dispatch]
```
alecbcs: I'd suggest limiting the execution of CI to the `devel` branch (and to PRs that target devel) which will reduce the number of parallel executions that can occur when you push to a branch and then create a PR. Additionally helps reduce the load on the limited number of Actions Runners we all share in the LLNL org.

https://github.com/LLNL/Spindle/blob/d5e582f316e03f23afcf6f410b06566ebade5f3c/.github/workflows/ci.yml#L2-L8

.github/workflows/CI.yml
```diff
@@ -0,0 +1,97 @@
...
+
+jobs:
+  spindle-serial-ubuntu:
+    name: Testsuite (Serial, Ubuntu)
+    environment: Spindle CI
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Check out Spindle
+        uses: actions/checkout@v5
```
alecbcs: To minimize the chance of external modifications creating a failure state in CI, I typically recommend that GitHub projects lock their actions to a specific commit and then use GitHub Dependabot to automatically bump that commit periodically.

https://github.com/LLNL/Spindle/blob/d5e582f316e03f23afcf6f410b06566ebade5f3c/.github/workflows/style.yml#L12

https://github.com/LLNL/Spindle/blob/d5e582f316e03f23afcf6f410b06566ebade5f3c/.github/dependabot.yml#L1-L6

.github/workflows/CI.yml
```diff
@@ -0,0 +1,97 @@
...
+        with:
+          path: 'Spindle'
```
alecbcs:
```suggestion
```

Did you all decide to clone into `Spindle/` for a particular use case? Normally we recommend to clone without a path which clone into the CI/CD runner current working directory. (And then minimizing the number of cd hierarchies you need later on.)

alecbcs (commented):
Hi @nchaimov! Nice job on the CI/CD pipeline! A couple quick suggestions and otherwise this looks good to me. 

- Tiny knitpick to follow common standards I'd suggest changing `CI.yml` -> `ci.yml`

I submitted a simple example CI/CD workflow in #106 to show the best practices I typically build off of which allow for filtering based on files changed, and break the workflows out into their own separate files.

.github/workflows/CI.yml
nchaimov: I will do this, thanks

.github/workflows/CI.yml
nchaimov: I don't think this is needed for our use cases. I can get rid of it.

nchaimov:
I've made the following changes: 

- renamed from `CI.yml` to `ci.yml`
- removed checkout path
- use specific commit for actions
- use Dependabot to update actions
- only run on commits or pull requests to `devel`
- added `permissions` and `concurrency` sections from #106

mplegendre:
I think this is looking great. Anything else before I merge?

nchaimov:
It's ready from my perspective

alecbcs (approved):
Looks good to me. Thanks @nchaimov for the quick turnaround!

PR #103, opened 2025-09-24
Handshake security

yaelhoma:
Adds random number exchange to handshake protocol integrated with all sec-types.

### DIFF
--- a/src/cobo/handshake.c
+++ b/src/cobo/handshake.c
@@ -29,6 +29,7 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include <sys/time.h>
 #include <sys/types.h>
 #include <sys/socket.h>
+#include <sys/random.h>
 #include <netinet/ip.h>
 
 #include "handshake.h"
@@ -82,24 +83,36 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #define CLIENT_TO_SERVER_SIG 0x9B1CC028
 
 #define HSHAKE_AGAIN -16
+#define HOSTNAME_MAX_LEN 256
 
 typedef struct {
+   uint64_t random_number;
+   uint64_t session_id;
+   uint32_t counter;
+   char hostname[HOSTNAME_MAX_LEN];
+} random_number_t;
+
+typedef struct {
+   random_number_t random_number;
    uint32_t signature;
    uint16_t server_port;
    uint16_t client_port;
    uid_t uid;
    gid_t gid;
-   uint64_t session_id;
    unsigned char server_addr[MAX_ADDR_LEN];
    unsigned char client_addr[MAX_ADDR_LEN];
 } handshake_packet_t;
+
+typedef struct {
+   random_number_t random_number;
+   uint32_t signature;
+} random_number_packet_t;
 
 typedef struct {
    int i_am_server;
    struct sockaddr server_addr;
    struct sockaddr client_addr;
 } connection_info_t;
-
 
 static FILE *debug_file = NULL;
 static char *last_error_message = NULL;
@@ -109,6 +122,7 @@ static char *saved_key_filepath;
 static unsigned int saved_key_len;
 static connection_info_t *saved_conninfo;
 static int timeout_seconds = 0;
+static uint32_t unique_number_counter = 0;
 
 /** Routines for creating a handshake_packet_t **/
 static int encode_addr(struct sockaddr *addr, unsigned char *target_addr, uint16_t *port);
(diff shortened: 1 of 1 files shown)


src/cobo/handshake.c
```diff
@@ -410,13 +465,66 @@ static int encode_addr(struct sockaddr *addr, unsigned char *target_addr, uint16
    return 0;
 }
 
+int random_number(uint64_t *unique_value) {
+   int local_errno;
+   uint64_t rand_num, uuid_num;
+
+   //random number file
+   FILE *random_file = fopen("/dev/random", "r");
```
mplegendre: Suggest using linux's getrandom() call (which wraps reading of /dev/urandom), which looks to be the preferred way to do this.

src/cobo/handshake.c
```diff
@@ -410,13 +465,66 @@ static int encode_addr(struct sockaddr *addr, unsigned char *target_addr, uint16
...
+
+   //uuid file
+   FILE *uuid_file = fopen("/proc/sys/kernel/random/uuid", "r");
+   if (!uuid_file) {
+      local_errno = errno;   
+      debug_printf("Failed to open /proc/sys/kernel/random/uuid. Error: %s\n",strerror(local_errno));
+      fclose(uuid_file);
+      return -1;
+   }
+   if (fread(&uuid_num, sizeof(uuid_num), 1, uuid_file) != 1) {
```
mplegendre: A) The /proc/sys/kernel/random/uuid file returns and ASCII encoded number, but this is reading it like a binary number. 
B) If we're doing a read/fread of random data we need to handle the EINTR return codes

src/cobo/handshake.c
```diff
@@ -410,13 +465,66 @@ static int encode_addr(struct sockaddr *addr, unsigned char *target_addr, uint16
...
+      debug_printf("Failed to read uuid. Error: %s\n",strerror(local_errno)); 
+      fclose(uuid_file);
+      return -1;
+   }
+   fclose(uuid_file);
+
+   time_t now = time(NULL);
+
+   // Combine timestamp,random number, uuid
+   *unique_value = now ^ rand_num ^ uuid_num;
```
mplegendre: We don't want to xor the data together. We're essentially xor'ing a random number with a partially-random number with a non-random number. I don't think that produces anything interesting. The goal is to make a unique number. I think we could do that by concat'ing into 128 bits (or fill in multiple 64-bit fields) with a random number, session id, and a counter.

mplegendre (changes requested):

src/cobo/handshake.c
```diff
@@ -109,38 +122,40 @@ static char *saved_key_filepath;
 static unsigned int saved_key_len;
 static connection_info_t *saved_conninfo;
 static int timeout_seconds = 0;
+uint32_t local_global_countr = 0;
```
mplegendre: Make this a static variable: 'static uint32_t local_global_countr'
That keeps it scope limited to this file.

Also suggest a more descriptive name for what it does, i.e., unique_number_counter

src/cobo/handshake.c
```diff
@@ -327,16 +344,65 @@ static int handshake_main(int sockfd, handshake_protocol_t *hdata, uint64_t sess
...
+      debug_printf("Problem receiving packet\n");
+      return_result = result;
+      socket_error = 1;
+      goto done;
+   }
+
+   /**
+    * Decrypt their packet with their random number
+   **/
+   debug_printf("Decrypting their random number packet\n");
```
mplegendre: Misleading comment and printf. This is decrypting the main handshake packet (which does contain the initial random number). But it is not the random number packet, which this comment implies.

src/cobo/handshake.c
```diff
@@ -998,25 +1073,29 @@ static int key_decrypt_packet(unsigned char *key, unsigned int key_len,
 #endif
 }
 
-static int compare_packets(handshake_packet_t *expected_packet,
-                           handshake_packet_t *recvd_packet)
+static int compare_packets(handshake_packet_t *expected_packet, handshake_packet_t *recvd_packet,
```
mplegendre: Let's break this into two functions: compare_handshake_packets() and compare_random_number_packets().
Then compare_handshake_packets() after the first exchange, but before doing the random packet exchange. 
Then compare_random_number_packets() after the random packet exchange.

Two reasons for this. A) It's a common occurrence for two different spindle sessions to connect to each other and realize they are the wrong sessions (the HSHAKE_DROP_CONNECTION test below). Let's get those dropped quickly before sending the other packets. B) No reason to be signing and sending those other packets when we can do an earlier test for common cases..

src/cobo/handshake.c
```diff
@@ -327,16 +344,65 @@ static int handshake_main(int sockfd, handshake_protocol_t *hdata, uint64_t sess
...
+    * Decrypt their packet with their random number
+   **/
+   debug_printf("Decrypting their random number packet\n");
+   result = decrypt_packet(hdata, &recvd_packet, sizeof(handshake_packet_t), recvd_long_packet_buffer, recvd_long_packet_buffer_size);
+   if (result < 0) {
+      debug_printf("Error decrypting and checking received packet\n");
+      return_result = result;
+      goto done;
+   }
+
```
mplegendre, line 366: See comment below about splitting up compare_packet(). The compare_handshake_packet() would go here.

src/cobo/handshake.c
```diff
@@ -327,16 +344,65 @@ static int handshake_main(int sockfd, handshake_protocol_t *hdata, uint64_t sess
...
+   //encrypt random number packet that they are expecting
+   random_number_packet.random_number = recvd_packet.random_number;
+   random_number_packet.signature = packet.signature;
+   result = encrypt_packet(hdata, &random_number_packet, sizeof(random_number_packet_t), &packet_buffer, &packet_buffer_size);
```
mplegendre, line 394: There's a memory leak on packet_buffer here. Before this commit packet_buffer was allocated and free'd once at the exit to this function. Now it's allocated twice and free'd once at the end of the function. It needs to be free'd before being passed to encryp_packet.

src/cobo/handshake.c
```diff
@@ -410,13 +473,52 @@ static int encode_addr(struct sockaddr *addr, unsigned char *target_addr, uint16
...
+      debug_printf("Trouble getting hostname. Error: %s\n",strerror(local_errno));
+      return -1;
+   }
+   hostname_buf[HOSTNAME_MAX_LEN-1] = '\0';
+   
+   //random number file
+   result = getrandom((void*) &rand_num->random_number,sizeof(rand_num->random_number),GRND_RANDOM|GRND_NONBLOCK); 
+   if (result<0) {
+      local_errno = errno;
+      if (local_errno == EAGAIN){
```
mplegendre: Need to handle EINTR here.

Also we don't want EAGAIN to be a fatal error (nor EINTR).

mplegendre (changes requested):

PR #108, opened 2025-10-03
Cachepath work broken out into functional commits

### LINKED ISSUE #61: Allow specification for cache, daemon, and fifo paths
Intended for testing "rabbit" rack storage on LLNL's Tuolumne supercomputer.

rountree:
Fixes #61


src/client/client/intercept_exec.c
```diff
@@ -198,13 +196,12 @@ static char **updateEnvironment(char **envp, int *num_modified, int propogate_sp
    if (envp) {
       debug_printf2("Propogating spindle environment by copying it to new envp list\n");
       for (cur = (char **) envp; *cur; cur++, orig_size++);
-      new_size = orig_size + 10;
+      new_size = orig_size + 20;
```
mplegendre: Should be 'new_size = orig_size + 9'. Eight environment variables get propogated, plus one slot for the NULL.

src/client/client/intercept_readlink.c
```diff
@@ -31,19 +31,25 @@ Place, Suite 330, Boston, MA 02111-1307 USA
...
 ssize_t (*orig_readlinkat)(int dirfd, const char *pathname, char *buf, size_t bufsiz);
 
-extern char *location;
+static char *cachepath;
+
+void set_intercept_readlink_cachepath( char *chosen_realized_cachepath, char *chosen_parsed_cachepath, char *chosen_symbolic_cachepath ){
+    cachepath = chosen_realized_cachepath;
+    chosen_parsed_cachepath = chosen_parsed_cachepath;
+    chosen_symbolic_cachepath = chosen_symbolic_cachepath;
+}
```
mplegendre: I'm a bit confused by the set_intercept_readlink_cachepath() and set_should_intercept_cachepath(). Seems like we're keeping multiple copies of the paths in different static variables, then calling each of these functions to set the different copies.

Different variables holding the same copies of information are a source of bugs. Let's consolidate the variables holding paths in the clients to just one instance.

Or am I missing something where these are different?

src/server/auditserver/ldcs_audit_server_process.h
```diff
@@ -126,6 +126,11 @@ struct ldcs_process_data_struct
   ldcs_dist_model_t dist_model;
   ldcs_client_t* client_table;
   char *location;
```
mplegendre: We've still got 'location'. What's does 'location' represent anymore? Is it a duplicate variable for cachepath?

Also, please comment what these variables contain.

src/include/spindle_launch.h
```diff
@@ -126,6 +126,11 @@ typedef struct {
    /* The local-disk location where Spindle will store its cache */
    char *location;
 
+   /* Path[s] for cached libraries. */
+   char *candidate_cachepaths;  /* Colon-separated list of candidate paths (max 64) */
+   char *chosen_cachepath;      /* The consensus path (same across all nodes). */
+   uint64_t cachepath_bitidx;   /* Bit index used by allReduce() to arrive at consensus. */
+
```
mplegendre, line 131: spindle_launch is the user interface function. We don't need to expose chosen_cachepath or cachepath_bitidx to the user. They're not selecting or reading those values.

Also, what's the difference between 'location' and 'candidate_cachepaths' here? What's it mean if a user sets both or only one?

src/client/client_comlib/client_api.c
```diff
@@ -37,6 +37,59 @@ static struct lock_t comm_lock;
...
+   if (message.header.type != LDCS_MSG_CHOSEN_CACHEPATH || message.header.len > MAX_PATH_LEN) {
+      err_printf("Got unexpected message of type %d\n", (int) message.header.type);
+      assert(0);
+   }
+   if( chosen_realized_cachepath ){
+       *chosen_realized_cachepath = strdup( buffer );
+   }
+
+   COMM_LOCK;
+   client_recv_msg_static(fd, &message, LDCS_READ_BLOCK);
```
mplegendre: Kind of messy doing one send and three recvs. In other parts of spindle the recv would be one message with three strings rather than three messages.

src/client/client_comlib/client_api.c
```diff
@@ -37,6 +37,59 @@ static struct lock_t comm_lock;
...
+      assert(0);
+   }
+   if( chosen_parsed_cachepath ){
+       *chosen_parsed_cachepath = strdup( buffer );
+   }
+
+   COMM_LOCK;
+   client_recv_msg_static(fd, &message, LDCS_READ_BLOCK);
+   COMM_UNLOCK;
+
```
mplegendre: If we keep this at one-send/three-recvs, then the locking here isn't correct. We're guarding against other threads in the same application process triggering spindle operations in parallel. This structure would allow another thread to grab the lock between the send/recvs and then recv the message meant for this thread.

Keep the lock through the entire send/recv stream (or condense this down to one recv message locked with the one send).

src/client/client_comlib/client_api.c
```diff
@@ -37,6 +37,59 @@ static struct lock_t comm_lock;
...
+       *chosen_realized_cachepath = strdup( buffer );
+   }
+
+   COMM_LOCK;
+   client_recv_msg_static(fd, &message, LDCS_READ_BLOCK);
+   COMM_UNLOCK;
+
+   if (message.header.type != LDCS_MSG_CHOSEN_CACHEPATH || message.header.len > MAX_PATH_LEN) {
+      err_printf("Got unexpected message of type %d\n", (int) message.header.type);
+      assert(0);
```
mplegendre: As per our other commits, move away from assert(0) on network failure.

src/server/auditserver/ldcs_audit_server_handlers.c
```diff
@@ -2936,6 +2943,79 @@ static int handle_client_pickone_msg(ldcs_process_data_t *procdata, int nc, ldcs
...
+static int handle_cachepath_consensus(ldcs_process_data_t *procdata, ldcs_message_t *msg){
+
+    int num_children = ldcs_audit_server_md_get_num_children(procdata);
+
+    if (num_children) {
+        spindle_broadcast(procdata, msg);
+        msgbundle_force_flush(procdata);
+    }
+
+    ldcs_audit_server_md_consensus(procdata, msg);
```
mplegendre: Suggest renaming ldcs_audit_server_md_consensus(...) to something like ldcs_audit_server_md_allreduce(MD_AND).

The ldcs_audit_server_md_* are all network operation focused, not spindle algorithm focused. That's the layer where you'd add a different network implementation (like infiniband), and we want to keep higher-level spindle concepts out of that layer.

src/server/auditserver/ldcs_audit_server_handlers.c
```diff
@@ -2936,6 +2943,79 @@ static int handle_client_pickone_msg(ldcs_process_data_t *procdata, int nc, ldcs
...
+   ldcs_send_msg(connid, &msg);
+   procdata->server_stat.clientmsg.cnt++;
+   procdata->server_stat.clientmsg.time += ldcs_get_time() - client->query_arrival_time;
+
+   msg.header.len = strlen(procdata->symbolic_cachepath) + 1;
+   msg.data = procdata->symbolic_cachepath;
+   ldcs_send_msg(connid, &msg);
+   procdata->server_stat.clientmsg.cnt++;
+   procdata->server_stat.clientmsg.time += ldcs_get_time() - client->query_arrival_time;
+
```
mplegendre, line 3005: The repeated adds to clientmsg.cnt and clientmsg.time is triple counting here.

src/flux/flux-spindle.c
```diff
@@ -416,7 +416,8 @@ static int sp_getopts (flux_shell_t *shell, struct spindle_ctx *ctx)
                         "numa", &numa,
```
mplegendre, line 416: Similar comment to previous, what's the difference between location and cachepaths now?

src/fe/startup/config_mgr.h
```diff
@@ -125,7 +126,8 @@ enum CmdlineShortOptions {
    shortSpindleLevel = 296,
    shortLocalPrefix = 297,
    shortExecExcludes = 298,
-   shortPatchLdso
+   shortPatchLdso,
+   shortCachePaths,
```
mplegendre: Just to keep the pattern going, could you add numbers to the enum values (and fix shortPatchLdso while at it). I know it's not necessary, but it makes it easier to look values from debug_printfs if they're explicit here.

src/cobo/cobo.c
```diff
@@ -1432,6 +1432,48 @@ int cobo_allgather_str(char* sendstr, char*** recvstr, char** recvbuf)
...
+                break;
+        }
+    }
+
+    /* forward data to parent if we're not rank 0, otherwise set the recvbuf */
+    if (cobo_me != 0) {
+        /* not the root, so forward our reduction result to our parent */
+        if (cobo_write_fd(cobo_parent_fd, pval, sizeof(*pval)) < 0) {
+            err_printf("Sending reduced data to parent failed\n");
+            exit(1);
```
mplegendre: Don't exit on network failure

mplegendre (changes requested):

src/client/client/intercept_exec.c
rountree: Fixed.

src/client/client/intercept_readlink.c
rountree: Two issues here:  
1) We're using setter functions to pass around pointers to static file-scoped variables.  Making the original variable global gets rid of the copies, but creates its own source of bugs.

2) We're keeping track of the realized, parsed, and symbolic versions of the cachepaths, Parsed and symbolic aren't used in `set_inercept_readlink_cachepath()` and symbolic isn't used in `set_should_intercept_cachepath()`.  (The code that references them in the setter function is there to silence the compiler warning.)

Let's err on the side of deleting more code.  Removed the two `set_*()` functions.

src/server/auditserver/ldcs_audit_server_process.h
rountree: The behavior of `location` hasn't changed, except that it's not longer used for cachepaths (it only affects the path used for fifo, the daemon, etc.).

We had discussed removing it entirely.  Let's leave that for a separate PR.

I've added comments.

src/include/spindle_launch.h
rountree: As above, `location` is used for paths to the fifo directory, the daemon file, and anything else that's not a cachepath.  I expect we'll remove it in a following PR.

Removing `chosen_cachepath` and `cachepath_bitidx` resulted in a modest string of perturbations, ending with relocating "parseloc.h" to `/src/util`.

src/client/client_comlib/client_api.c
rountree: Modified to use a single reply.  Removed future-proofing (not using the symbolic version of cachepaths).

src/client/client_comlib/client_api.c
rountree: Condensed to a single message.

src/client/client_comlib/client_api.c
rountree: Converted `assert(0)` to `return -1` in eight functions.

src/server/auditserver/ldcs_audit_server_handlers.c
rountree: Renamed/redefined to `ldcs_audit_server_md_allreduce_AND( uint64_t *val )`.

I wanted to use your `MD_AND` suggestion, but that's going to result in a duplicated op enum in md-space and coho-space. As we're only using the bitwise-and op in md-space, let's kick that problem down the road and put the op name in the function name.  Ugly, but effective.

src/server/auditserver/ldcs_audit_server_handlers.c
rountree: Resolved in earlier commit when we went to a single-message response.

src/flux/flux-spindle.c
rountree: "location" will be renamed to "commpath" in a subsequent PR.

src/fe/startup/config_mgr.h
rountree: Done.

src/cobo/cobo.c
rountree: Changed 27 instances of exiting on network failure to "return -1" and two to "return NULL".

Not confident that's a net gain.  If we end up pulling the network code out into its own library, we can revisit error handling then.

src/cobo/cobo.c
```diff
@@ -160,7 +160,7 @@ static char* cobo_getenv(char* envvar, int type)
     char* str = getenv(envvar);
     if (str == NULL && type == ENV_REQUIRED) {
         err_printf("Missing required environment variable: %s\n", envvar);
-        exit(1);
```
mplegendre, line 163: Was this commit intended to be part of the cachepath work? It looks more to be part of the network reliability work.

And if there's a reason to do this in the cachepath PR, we would need to take a more robust pass to make sure the various error returns don't just lead to crashes.

mplegendre: Oh, I see it's in response to my above comment. 

I'd more meant that we shouldn't propogate, not that clean them all up. We can move forward with this in the PR, but we'll want to come back and work through some of these new error return paths.

testsuite/test_driver.c
```diff
@@ -1142,27 +1142,6 @@ static char* getCacheLocation(char *env_var)
    return strdup(last_slash);
 }
 
-static int checkLinkForLeak(const char *path, const char *spindle_loc)
```
mplegendre, line 1145: This shouldn't be deleted. It's checking that spindle's readlink interception is working and that we don't have symlinks from /proc/PID/... pointing into the spindle cache location that can be seen by the application (which has happened before). 

We may need another way to get the cache path into the test, but we should keep the test.

mplegendre (commented):
Looks like the CI is failing on this PR due to unset TMPDIR changes. Resolve that and the deletion in test_driver.c and I think we'll be ready to merge.

src/cobo/cobo.c
rountree: Sounds good.

testsuite/test_driver.c
rountree: The cache path is available to the test, but the fifos are no longer in that path, so those files leak.  If I restore that code, these are the kind of errors I get on the first of the runTests:
```
 Error - [/p/vast1/rountree/repos/Spindle/testsuite/test_driver.c:1159] - Link at '/proc/self/fd/315' has path '/tmp/rountree/spindle.9d2b000000/spindle_comm/fifo-1379774-0', which leaks spindle path with 'spindle.9d2b000000'
 ```
I'll restore the `checkLinkForLeak` in the commpath PR and make sure that path gets into the `TestVerifier` class via another `<internal>` message (unless you'd rather I merge the commpath PR into this PR).  Just to be clear, the above file should NOT trigger an error, correct?

rountree: Well, that was a good solution to a problem we don't have yet.
Here's the actual problem:

https://github.com/LLNL/Spindle/blob/d8ac4c58a0f2b4db00f4e420ccb7e608489bc687/testsuite/test_driver.c#L1241

Should be a bit simpler to solve.

Issue #114, opened 2025-10-10
Clients should tolerate disconnect due to server faults


rountree:
Discussed as part of the network resilience efforts.

PR #115, opened 2025-10-10
Adds disconnect handling to client comms.

rountree:
Addresses #114.

In failure cases where the client can determine a disconnect occurred, it should fall back to non-Spindle operation and continue.

Code copied from Matt's earlier resilience work.


rountree:
Not ready for review.  Hitting asserts...

PR #117, opened 2025-10-24
Commpath

rountree:
Renames `location` environment variables, fields, and enums to `commpath`.

Builds with --with-rm=[flux|serial|slurm].  Can't build slurm-plugin, but that's not due to these changes.


rountree:
Found a failure case.  Hold off on review for now.

configure
```diff
@@ -846,7 +846,8 @@ enable_libtool_lock
 enable_maintainer_mode
 with_default_port
 with_default_num_ports
-with_localstorage
```
nchaimov, line 850: The configure script will give a warning for unrecognized arguments, so removing the `--with-localstorage` causes builds that still specify `--with-localstorage` to configure without error, but the option is unused. For example, the script in the CI environment that builds Spindle uses `--with-localstorage=/tmp`, and you can see the warning in the CI output: 

```
#18 0.165 configure: WARNING: unrecognized options: --with-localstorage
```
https://github.com/llnl/Spindle/actions/runs/21076617750/job/60620001124?pr=117#step:4:826

I wonder if it would make sense to either:

1. Give an error rather than a warning if `--with-localstorage` is specified, where the error message says that `--with-cachepaths` and `--with-commpath` should be used instead; or,
2. Still accept `--with-localstorage` and use it to set the cachepaths and commpath value.

We should also change the Spindle configure line used in CI to use the new arguments instead of `--with-localstorage`, in `containers/spindle-serial-ubuntu/scripts/build_spindle.sh`, `containers/spindle-flux-ubuntu/scripts/build_spindle.sh`, and `containers/spindle-slurm-ubuntu/testing/scripts/build_spindle.sh`.

testsuite/test_driver.c
```diff
@@ -1284,34 +1283,39 @@ void check_for_path_leaks()
    for (d = readdir(proc_fds); d != NULL; d = readdir(proc_fds)) {
       if (d->d_name[0] == '.')
          continue;
+      // Ignore Spindle fifo files for now.
+      if ( strncmp( "315", d->d_name, 3 ) == 0 )
+          continue;
+      if ( strncmp( "316", d->d_name, 3 ) == 0 )
+          continue;
```
nchaimov, line 1290: Where do these numbers (315, 316) come from?

src/trivial_change
```diff
@@ -0,0 +1 @@
+Hello, world!
```
nchaimov: I think this was just added to force the CI to run and should be removed.

configure
rountree: Per Matt, map this to both cache and compaths, although those take precedence.

testsuite/test_driver.c
rountree: A little above the safe zone for tcsh closings.

configure
rountree: I think the best way forward is removing `--with-localstorage`, forcing a configure error if it's used, and updating the CI configure scripts.  The semantics for `--with-localstorage` are slightly different from `--with-cachepaths` in that the latter can accept multiple colon-separated paths, and communicating to the user which option takes precedence and why appears to be more trouble than it's worth for an option that's not widely used.

The output of `configure --help` now looks like this:
```
  --with-localstorage=DIR (obsolete)
                          Use --with-cachepaths and --with-commpath instead.
```

The configure error message looks like this:
```
$ ./configure --with_localstorage=foo
configure: error: unrecognized option: `--with-localstorage=foo'
...
configure: error: requested obsolete option --with-localstorage.  Use --with-cachepaths and --with-commpath instead.
```

PR #120, opened 2025-11-21
Fixes/updates for compilation under gcc 15.1.0 -Wall -Wextra -Werror 

### LINKED ISSUE #94: Clean up warnings

rountree:
Fixes and updates to compile under gcc 15.1.0 with the flags `-Wall`, `-Wextra`, and `-Werror`.

Commits are based on the class of errors/warnings rather than files.

```
commit 47cb39:  Fixes three errors caused by stricter gcc defaults.
commit c6f85d:  Fixes one [-Wdiscarded-qualifiers] and one [-Wstringop-truncation].
commit a7260c:  Fixes fallthrough and initialization warnings.
commit acc263:  Fixes -Wsign-compare warnings.
commit 5182e8:  Fixes -Wformat warnings.
commit 32d0c6:  Suppresses -Wunused-parameter warnings by casting to void.
```
Fixes #94

### DIFF
--- a/src/biter/shmutil.h
+++ b/src/biter/shmutil.h
@@ -85,6 +85,6 @@ int setup_ids(shminfo_t *shminfo);
 int take_heap_lock(shminfo_t *shminfo);
 int release_heap_lock(shminfo_t *shminfo);
 
-void update_shm_id();
+void update_shm_id(shminfo_t *shminfo);
 
 #endif
--- a/src/biter/shmutil.h
+++ b/src/biter/shmutil.h
@@ -17,18 +17,20 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #if !defined(SHMUTIL_H_)
 #define SHMUTIL_H_
 
+#include <sys/types.h>
 #include "sheep.h"
 #include "spindle_launch.h"
 
 typedef struct {
    volatile unsigned long *lock;
-   volatile unsigned long *held_by;
+   volatile pid_t *held_by;
    int ref_count;
 } lock_t;
 
 typedef struct {
    volatile int cur_id;
-   unsigned long locks[2];
+   unsigned long *lock;
+   pid_t *held_by;
 } base_header_t;
 
 #include "demultiplex.h"
@@ -43,11 +45,13 @@ typedef struct {
    int num_ranks;
    int read_file;
    int num_started;
-   unsigned long locks[6];
+   unsigned long lock[3];
+   pid_t held_by[3];
 } biter_header_t;
 
 typedef struct {
-   unsigned long locks[2];
+   unsigned long *lock;
+   pid_t *held_by;
    sheep_ptr_t hash;
    sheep_ptr_t lru_head;
    sheep_ptr_t lru_end;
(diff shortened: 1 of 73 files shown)


mplegendre:
This looks like it needs a rebase. Existing commits from both Nick and I are in here with changed hashes.
How does this relate to your other warning cleanup?

rountree:
I closed out the other warning cleanup PR; this PR replaces it.  I'll get a rebase committed later this morning.

rountree:
Just a single file needed cleanup.  Compiles without errors or warning with `-Wall -Wextra -Werror` using gcc-15.1.0 on Tuolumne.  All runTests passed.

src/biter/shmutil.h
```diff
@@ -17,18 +17,20 @@ Place, Suite 330, Boston, MA 02111-1307 USA
...
 typedef struct {
    volatile unsigned long *lock;
-   volatile unsigned long *held_by;
+   volatile pid_t *held_by;
    int ref_count;
 } lock_t;
 
 typedef struct {
    volatile int cur_id;
-   unsigned long locks[2];
```
mplegendre, line 31: I'm nervous about changing the locks[] layout here. I don't remember why I wrote it this way, but I wouldn't have done something this odd without reason. Given that arrays and structs can have different packing (even if on x86_64 it shouldn't pack struct vs array different here), it might have had something to do with that. There's also some memory barriers around the code that uses these arrays, and there's always tough subtleties when those are involved.

That said, I can't re-determine why these should be arrays with just code inspection. And it's too bad I didn't comment why it's like this.

Given that this is a warning cleanup commit, let's not risk subtle semantic changes.

(And this applies to the two other array->struct conversions below as well)

src/client/client/realpath.c
```diff
@@ -109,6 +109,7 @@ B. Use spindle versions of the IO accessing lower-level routines.
 The source file is glibc/stdlib/canonicalize.c at git commit 1894e219dc530d7074085e95ffe3c1e66cebc072
  */
 
+#pragma GCC diagnostic ignored "-Wsign-compare"
```
mplegendre: There are macros in src/utils/ccwarns.h for doing these

mplegendre:
Also, it looks like you did a merge to resolve the conflicts. 
But it needs a rebase that adds only the warning commits. If you look at the commit list for the PR, there's several other commits that are already merged (but have different hashes in this PR).

src/client/client/realpath.c
rountree: Those are handy.  I've narrowed the range of the disabled warning to a single `while` loop.  Putting the macro around the `while` test seems safe, but I'm not familiar with the rules of putting `#praga` within statements or across loop boundaries.

src/biter/shmutil.h
rountree: Fixed.  Getting the `#include`s to work required updating a couple of `Makefile.am`s.

rountree:
I just noticed that `-Wall -Wextra -Werror` is causing failures in the configure process.  These don't affect the results of `runTests`, but we might want to get those cleaned up.

Issue #139, opened 2026-02-25
DATA_GOT_TYPE undeclared on non-x86_64 architectures (ppc64, ppc64le, aarch64)


SamHerts:
### Building Spindle on non-x86_64 architectures (aarch64, ppc64, ppc64le) fails with a compile error
```
writablegot.c:124:50: error: 'DATA_GOT_TYPE' undeclared (first use in this function)
  124 |          if (ELF64_R_TYPE(datarels[j].r_info) != DATA_GOT_TYPE)
      |                                                  ^~~~~~~~~~~~~
```

### Root cause:

DATA_GOT_TYPE is defined in the architecture guard block in

https://github.com/llnl/Spindle/blob/4174224176e0840e0f0ed3fffdb16115a88e4a4b/src/client/auditclient/writablegot.c#L47-L50

but not defined for other architectures 
https://github.com/llnl/Spindle/blob/4174224176e0840e0f0ed3fffdb16115a88e4a4b/src/client/auditclient/writablegot.c#L51-L53
https://github.com/llnl/Spindle/blob/4174224176e0840e0f0ed3fffdb16115a88e4a4b/src/client/auditclient/writablegot.c#L54-L56

The macro is then used unconditionally at line 124 regardless of architecture, causing the build to
fail on any architecture other than x86_64.

https://github.com/llnl/Spindle/blob/4174224176e0840e0f0ed3fffdb16115a88e4a4b/src/client/auditclient/writablegot.c#L123-L125

Expected behavior

Spindle should build successfully on all supported architectures (x86_64, ppc64, aarch64).

Suggested fix

Add the appropriate ELF GLOB_DAT relocation type constant for each architecture in the guard block. The
 correct constant names should be confirmed from <elf.h> on each target platform
| Arch | VAR_NAME | Source
|----|----|----|
| PPC64 | R_PPC64_GLOB_DAT | [elf.h L2710](https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/elf.h;h=46a01281cb0fb5322d5124f0443c11dea4d5b721;hb=HEAD#l2710) |
| Aarch64 | R_AARCH64_GLOB_DAT | [elf.h L3026](https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/elf.h;h=46a01281cb0fb5322d5124f0443c11dea4d5b721;hb=HEAD#l3026)

Environment

- Spindle: devel branch (commit 4174224)
[...]

PR #140, opened 2026-02-25
Aarch64/Arm64 compatibility fixes for #139, #141, #142

SamHerts:
Fixes issue #139

Fixes issue #141 

Fixes issue #142

### DIFF
--- a/src/client/auditclient/writablegot.c
+++ b/src/client/auditclient/writablegot.c
@@ -51,9 +51,11 @@ static signed int gots_size;
 #elif defined(arch_ppc64) || defined(arch_ppc64le)
 #define DEFAULT_RELENTSZ 24
 #define EXTRA_GOT_ENTRIES 2
+#define DATA_GOT_TYPE R_PPC64_GLOB_DAT
 #elif defined(arch_aarch64)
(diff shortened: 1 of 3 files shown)


mplegendre:
I've no objections to the code changes. But do you have needs for spindle on either ppc64le or aarch64? 

I was intending to let the ppc64le port wilt, as I no longer have easy testing on ppc64le, nor do I know of big ppc64le systems that require it.

Spindle's aarch64 port is essentially abandoned. There's a serious bug in GLIBC on aarch64 that leads to FP register corruption when using LD_AUDIT (which spindle would trigger). I could dig into the status of that bug, but last I check there wasn't a path to fixing it.

SamHerts:
> I was intending to let the ppc64le port wilt, as I no longer have easy testing on ppc64le, nor do I know of big ppc64le systems that require it.

For ppc64, I have no need whatsoever. I just added it since the definition was available in the elf.h source.

As for aarch64, I'd definitely like to be able to run Spindle on the architecture, since some systems in ORNL's testbed are aarch64. 

Can you provide some links for information regarding the GLIBC issue?

PR #148, opened 2026-03-18
Fixes for spindle's internal path parsing

mplegendre:
This fixes two bugs, both related to the spindle server's path parsing:

First, files in the root filesystem were being mis-parsed by internal spindle path parsing. Something like "/bin" would get parsed into a file and directory component, and we were incorrectly making the directory the empty string in this case. That would fool other parts of spindle into thinking it was a relative path, and thus "/bin" would get cached by spindle as "${PWD}/bin".

While fixing that I added additional tests around path parsing and discovered more issues around symlinks and paths that use "..". In short, we were resolving the ".." in those paths before resolving symlinks. That's not wise, as a ".." in a path moves up a directory from a symlink target, not source.

I significantly rewrote path parsing for a couple places in spindle where this matters, and added new tests.


mplegendre:
@rountree -- Ping. Can you review this?

nchaimov:
I've looked at the path parsing code and done some testing with the `parseFilenameNoAlloc2` function. I found that if we give `parseFilenameNoAlloc2` a path like `"/bin/.."`, then we end up with `dir = "/bin"` and `file = ".."`, even though in this case the whole path refers to a directory, not a file. I know that it's intentionally not handling `..` components, but is it acceptable for `".."` to end up as the `file`?

The static TLS commit is labeled as a WIP — was that meant to be included in this PR?

mplegendre:
Good catch on the "/bin/..". That's an annoying case, and I'll have to think about what we do there.
The WIP commit slipped in accidently. Thanks for catching that too.

Will fix.

rountree:
Adding to Nick's observation:  `/bin/../..` and `/bin/../../` should be test cases as well.

No issues compiling w/ gcc 15.1.0 with `-Wall -Wextra -Werror`.

mplegendre:
It's not off of /bin, but the test on line 1109 (the one that adds the "slashes" variable) puts enough "../" on the prefix of a path to go two past the root directory. Then adds enough path back to get to a real file and test that spindle got there correctly.

The catch with testing actual "/bin" is that it's not accessing files that spindle would relocate, and we can't guarantee what files are in there.

mplegendre:
I went over the /bin/.. case in detail, and it looks like we handle it okay. There's definitely some oddities with spindle putting ".." into an internal variable named 'file'. But we ultimately always supported spindle operations on directories (even if that wasn't common), and ".." just gets treated like other directories and handled correctly.

nchaimov:
That was the only issue I found, so I think this is ready as far as I'm concerned.

nchaimov (approved):

PR #150, opened 2026-03-23
Have la_version return provide audit version rather than hardcoded 1

### LINKED ISSUE #143: la_version() returns hardcoded 1, incompatible with glibc ≥ 2.35 audit interface version 2
Spindle's audit libraries (libspindle_audit_pipe.so and libspindle_subaudit_pipe.so) return a hardcoded value of 1 from la_version(). Glibc 2.35 (released February 2022) bumped LAV_CURRENT to 2. On platforms where glibc enforces a minimum accepted version of 2, Spindle's audit libraries are silently rejected at
load time with the following error:

```
ERROR: audit interface 'libspindle_audit_pipe.so' requires version 1
       (maximum supported version 2); ignored.
```

When the audit library is ignored, all symbol interception fails and any spawned process that depends on LD_AUDIT pointing to a Spindle library hangs indefinitely waiting for server communication that never
arrives, causing test timeouts.

Affected platforms

This is currently observed on Ubuntu 24.04 arm64 (aarch64). Ubuntu 24.04 x86_64 is not affected at this time, believed to be due to a difference in patch level between the amd64 and arm64 libc6 packages within
Ubuntu 24.04 Noble.

Root cause

spindle_la_version() is implemented in two places, both returning 1 unconditionally and discarding the version argument that glibc passes:

https://github.com/llnl/Spindle/blob/e731f244db5e8f42c4fcd01abb4673d3c0da93c3/src/client/auditclient/auditclient.c#L30-L35

https://github.com/llnl/Spindle/blob/e731f244db5e8f42c4fcd01abb4673d3c0da93c3/src/client/subaudit/subaudit.c#L28-L43

Additionally, the return -1 error path in subaudit.c is typed as unsigned int, so it silently returns 0xFFFFFFFF rather than -1, which would also be rejected by glibc (lav > LAV_CURRENT). Replacing with 0 would automatically disable the audit interface:

```
if (lav == 0)
    ...
    unload_audit_module (dlmargs.map, original_tls_idx);
    return;
}
```
[GLibC 2.35 Source Reference](https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/rtld.c;h=c81542a08d619d90e0d15270664bbd3cdc65522f;hb=refs/heads/release/2.35/master#l994)

Fix

  In both spindle_la_version implementations, replace return 1 with return version, and remove the (void)version suppression. The error path in subaudit.c should return 0 (which glibc treats as an explicit
  rejection) rather than -1.

auditclient.c:
```
unsigned int spindle_la_version(unsigned int version) 
 { 
    patchDTV_init(); 
    return version; 
 } 
```

subaudit.c:
```
unsigned int spindle_la_version(unsigned int version) 
 { 
    int result; 
    int binding_offset = 0; 
  
    result = get_ldso_metadata_bindingoffset(&binding_offset); 
    if (result == -1) { 
       err_printf("Unable to lookup binding offset\n"); 
       return 0; 
    } 
    debug_printf3("Updating subaudit bindings with offset %d\n", binding_offset); 
    init_plt_binding_func(binding_offset); 
  
    return version; 
 } 
```

Reference

  - [glibc 2.35 release (February 2022)](https://lists.gnu.org/archive/html/info-gnu/2022-02/msg00002.html#:~:text=The%20GNU%20C%20Library%20version%202.35%20is%20now%20available%20The,%2C%20r_next%2C%20support%20multiple%20namespaces.)

mplegendre:
Fixes #143 

Change the compatibility version returned to AUDIT via la_version() to reflect whichever version is actually running. 

@SamHerts -- I don't have an aarch64 system to test this against. Can you let us know if this works?

### DIFF
--- a/src/client/auditclient/auditclient.c
+++ b/src/client/auditclient/auditclient.c
@@ -29,9 +29,8 @@ extern void restore_pathpatch();
 
 unsigned int spindle_la_version(unsigned int version)
 {
-   (void)version;
    patchDTV_init();
-   return 1;
+   return version;
 }
 
 void spindle_la_activity (uintptr_t *cookie, unsigned int flag)
--- a/src/client/auditclient/auditclient_common.c
+++ b/src/client/auditclient/auditclient_common.c
@@ -48,7 +48,11 @@ unsigned int la_version(unsigned int version)
    debug_printf("la_version function is loaded at %p\n", la_version);
    debug_printf3("la_version(): %d\n", version);
    init_bindings_hash();
-   return spindle_la_version(version);
+   
+   result = spindle_la_version(version);
+   if (result == -1)
+      return 0;
+   return result;
 }
 
 char *la_objsearch(const char *name, uintptr_t *cookie, unsigned int flag)
--- a/src/client/subaudit/subaudit.c
+++ b/src/client/subaudit/subaudit.c
@@ -27,7 +27,6 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 
 unsigned int spindle_la_version(unsigned int version)
 {
-   (void)version;
    int result;
    int binding_offset = 0;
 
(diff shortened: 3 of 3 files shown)


SamHerts:
Oops, didn't see this until after submitting my own. Check out #151 since that adds CI/CD and adds a fix for Arm, but also refactors the auditclient to simpify the logic.

PR #151, opened 2026-03-24
Fix Arm64 support and add Arm64 to CI/CD tests

SamHerts:


mplegendre:
Unfortunately, I don't have a convenient Linux/ARM system available right now. So I'll have to ask questions rather than test this myself.


The code you refactored from x86_64 to aarch64 is handling a small annoying corner case around data relocations because of limitations around la_symbind. Have you verified that those limitations also exist on aarch64?

Specifically, the refactored code is handling the use case where a program references a symbol of interested as data. E.g.,
```
int (*open_fptr)(const char *, int, ...) = open;
```
We'd like symbol to intercept the reference to the 'open' function so we can replace it with the spindle version. But on x86_64 la_symbind() doesn't return binding information when a function is referenced as data. So this code is iterating over the symbol tables, looking for these references, and manually binding them. It's a pain. So can you verify that we need the same capability on aarch64? Does la_symbind() have the same limitation on aarch64?


(Otherwise this PR looks fine at a glance--I'll do a more detailed review soon-ish).

SamHerts:
I don't believe there is a dramatic difference with la_symbind() between x86 and aarch with regard to binding info, so I think yes it would have the same limitations.

PR #177, opened 2026-05-11
Fix bugs in argument passing to scripts

mplegendre:
We had some failures in an application that came down to Spindle changing how the $1 in a bash script changed depending on whether it was passed in or out of spindle. This is related to #96, which was essentially the same issue, though this extends the fix into more corner cases. Specifically when argv[0] of the script does not match the argument passed to the first parameter of exec.

In addition, this adds a more comprehensive test case that tries to exec a set of scripts under a combinatorial set of circumstances (exec vs execve, argv[0] matches exec or not, path search vs relative vs absolute, and shell type).

### DIFF
--- a/src/client/client/exec_util.c
+++ b/src/client/client/exec_util.c
@@ -129,7 +129,7 @@ static int parse_interp_args(char *interp_line, char **interp_exec, char ***inte
 }
 
 
-int adjust_if_script(const char *orig_path, char *reloc_path, char **argv, char **interp_path, char ***new_argv)
+int adjust_if_script(const char *orig_path, char *reloc_path, char **argv, char **interp_path, char ***new_argv, char *found_from_pathsearch)
 {
    int result, fd, argc, interp_argc, i, j, errcode;
    char interpreter_line[MAX_PATH_LEN+1];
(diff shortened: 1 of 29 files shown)


nchaimov:
Sorry, was trying out Copilot to see if it could find issues that I couldn't, and it posted without my permission! Not trying that again...

mplegendre:
Both those are valid finds, and good ones. The extra readlink is debugging code I accidently left in and the stat one is accurate. I'll fix those both.

It got some details wrong. The readlink not infinite recursion, just a performance hit, but still valid.

nchaimov:
Regardless, it does look like Copilot found what I think may be a real problem, though: the wrapper for `ldso_lxstat`, to which a new fallback was added, falls back to plain `stat`. Is that intentional?

src/client/client/exec_util.c
```diff
@@ -273,6 +281,9 @@ int exec_pathsearch(int ldcsid, const char *orig_exec, char **reloc_exec, int *e
       get_relocated_file(ldcsid, newexec, 1, reloc_exec, errcode, NULL);
       debug_printf2("Exec search request returned %s -> %s\n", newexec, *reloc_exec ? *reloc_exec : "NULL");
       if (*reloc_exec) {
+         if (orig_file_abspath) {
+            *orig_file_abspath = strdup(newexec);
```
nchaimov: A question: the added code above in `adjust_if_script` uses `spindle_strdup()` (and replaces a `strdup()` with `spindle_strdup()`, but here `strdup()` is used. Under what circumstances should one be used vs the other?

testsuite/runTests_template
```diff
@@ -190,6 +190,9 @@ CHECK_RETCODE
 ./run_driver --serial ./interpreter_test
 CHECK_RETCODE
 
+./run_driver --serial ./exec_shell
```
nchaimov, line 193: Unlike the other tests, this one doesn't print "PASSED." when it passes. For consistency, we might want to have it do that, so that it doesn't look like it crashed without printing.

```
Running: ./run_driver --serial ./interpreter_test
PASSED.
Running: ./run_driver --serial ./exec_shell
ALL TESTS PASSED
```

src/client/client/exec_util.c
mplegendre: I'll move it to spindle_strdup() for consistency.

At one point I had a memory heap corruption checker integrated with the spindle, which required all of the client heap calls to go through wrappers (I had to do the interception manually since the library namespace issues kept spindle from working with regular tools).

mplegendre:
I addressed the identified issues. 
Thanks for the feedback on them.

nchaimov (approved):

PR #178, opened 2026-05-13
Use --external-launcher for Slurm launch; fallback to old behavior; add srun wrapper CI job

nchaimov:
This contains the commit from #58, which uses `--external-launcher` with the Slurm srun wrapper, with the following changes:

- Rebased on devel so that we have the CI tests in place.
- Falls back to old behavior for Slurm before 23.11 (when `--external-launcher` was added).
- Adds new CI job which builds Spindle with rm=slurm and without rshlaunch, in order to test the changed code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,7 +151,7 @@ jobs:
           docker compose down
 
   spindle-slurm-ubuntu:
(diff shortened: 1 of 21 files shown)


mplegendre (approved):
I'm not sure another CI set was necessary for just this mode. But since the work is already done, I'm not going to object either.

Otherwise looks good.

PR #180, opened 2026-05-15
Fix GitHub hang

### LINKED ISSUE #179: Startup hang in oversubscribed containerized environments
During Github Actions CI/CD testing, the first test occasionally hangs.  Nick reproduced this in his own Docker environment.  Bumping the number of containers up to 32 in Github more or less guarantees multiple ranks will hang.  No OOM activity observed in dmesg logs.  gdb backtraces isolated the problem to certain ranks not seeing the decorated `shell.init` message arriving in the event log.

rountree:
Fixes #179

    Startup sequence in flux-spindle.c could hang in
    oversubscribed environments.

    In sp_init() [flux-spindle.c], when the notification of the
    shell.init even being added to the guest.exec.eventlog is
    propagated to a rank whose reactor has not yet started,
    the message fails to be parsed once the reactor eventually
    starts.  Parsing would happen on the arrival of a subsequent
    message, but in Spindle's case, that never happens, and
    the rank never proceeds to wait_for_shell_init().

    This commit fixes the above by using synchronous KVS
    queries.

    Tested across 300+ iterations on github using 32 nodes
    running on ~2 cores.  Prior to the fix, test execution
    would rarely get past the first iteration without multiple
    ranks hanging.  With the fix applied, no hangs were
    observed in 300+ runs.

    This approach does introduce a dependence of all N ranks
    on rank 0, which could lead to substantial slowdown at
    scale.  Given that the race condition has not been observed
    in production machines, it might be best to have this
    approach be a runtime option for CI/CD environments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
 
       - name: Setup Docker Compose
-        uses: docker/setup-compose-action@364cc21a5de5b1ee4a7f5f9d3fa374ce0ccde746
+        uses: docker/setup-compose-action@8cccb8c14b6500aaffebff1aa49c502c34d2e5e6
         with:
           version: latest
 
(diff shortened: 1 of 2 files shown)


rountree:
Passes with `runTests.py` but not `runTests`.  Back to the drawing board.

PR #182, opened 2026-05-22
Commpath

rountree:
Passed all tests on my clone.  Let's see what it does over here.


src/client/beboot/spindle_bootstrap.c
```diff
@@ -108,14 +108,18 @@ static void setup_environment()
...
    snprintf(rankinfo_str, 256, "%d %d %d %d %d", ldcsid, rankinfo[0], rankinfo[1], rankinfo[2], rankinfo[3]);
-   
+
    char *connection_str = NULL;
    if (opts & OPT_RELOCAOUT) 
       connection_str = client_get_connection_string(ldcsid);
 
+   char *chosen_parsed_cachepath = NULL;
+   send_cachepath_query( ldcsid , NULL, &chosen_parsed_cachepath);
+   assert( chosen_parsed_cachepath );
```
mplegendre: Don't assert.  Instead do an err_printf and return -1.

src/client/client/client.c
```diff
@@ -254,20 +250,22 @@ static int init_server_connection()
...
+      send_location(ldcsid, commpath);
       send_rankinfo_query(ldcsid, rankinfo+0, rankinfo+1, rankinfo+2, rankinfo+3);
 #if defined(LIBNUMA)      
       if (opts & OPT_NUMA)
          send_cpu(ldcsid, get_cur_cpu());
 #endif
    }
-   
+   send_cachepath_query( ldcsid, &chosen_realized_cachepath, &chosen_parsed_cachepath );
+   assert( chosen_realized_cachepath );
```
mplegendre: Again, don't assert in the client. err_printf and return -1.

src/client/client/client.c
```diff
@@ -475,7 +473,8 @@ char *client_library_load(const char *name)
 
    char *orig_file_name = (char *) name;
    if (is_in_spindle_cache(name)) {
-      debug_printf2("Library %s is in spindle cache (%s). Translating request\n", name, location);
+      assert( chosen_realized_cachepath );
```
mplegendre: Don't assert

src/client/client/intercept_readlink.c
```diff
@@ -31,19 +31,20 @@ Place, Suite 330, Boston, MA 02111-1307 USA
...
 static int fix_local_readlink(char *buf, size_t bufsiz)
 {
    char spindle_id[32];
-   int location_len, result;
+   int cachepath_len, result;
    char tmp[MAX_PATH_LEN+1];
+   extern char *chosen_realized_cachepath;
 
-   location_len = strlen(location);   
+   assert( chosen_realized_cachepath );
```
mplegendre: Remove assert

src/client/client/should_intercept.c
```diff
@@ -30,21 +30,20 @@
...
 int is_in_spindle_cache(const char *pathname)
 {
-   static int location_size = 0;
-   static int orig_location_size = 0;
-   if (!location_size) {
-      location_size = strlen(location);
+   static int cachepath_size = 0;
+   static int orig_cachepath_size = 0;
+   extern char *chosen_realized_cachepath, *chosen_parsed_cachepath;
+   assert( chosen_realized_cachepath );
```
mplegendre: Don't assert

src/client/client_comlib/client_api.c
```diff
@@ -36,7 +37,56 @@ static struct lock_t comm_lock;
 
 #define COMM_LOCK do { if (lock(&comm_lock) == -1) return -1; } while (0)
 #define COMM_UNLOCK unlock(&comm_lock)
-   
+
+
+int send_cachepath_query( int fd, char **chosen_realized_cachepath, char **chosen_parsed_cachepath){
```
mplegendre, line 42: I'm not fond of this mechanism--timeouts and retries are brittle as a general rule. Ultimately we're doing synchronization here, and we can do that without a timeout.

But let's clean this up in a subsequent commit and not hold up commpath.

src/client/client_comlib/client_api.c
```diff
@@ -36,7 +37,56 @@ static struct lock_t comm_lock;
...
+       client_send_msg(fd, &message);
+       client_recv_msg_static(fd, &message, LDCS_READ_BLOCK);
+
+       COMM_UNLOCK;
+
+       if( message.header.type == LDCS_MSG_NO_CACHEPATH_CONSENSUS_YET ){
+           if( retries++ >= max_retries ){
+               break;
+           }
+           nanosleep( &one_second, NULL );
```
mplegendre: One second is too long. Suggest .1 seconds.

src/client/client_comlib/client_api.c
```diff
@@ -36,7 +37,56 @@ static struct lock_t comm_lock;
...
+   do{
+       message.header.type = LDCS_MSG_CHOSEN_CACHEPATH_REQUEST;
+       message.header.len = 0;
+       message.data = buffer;
+
+       COMM_LOCK;
+
+       debug_printf3("sending message of type: CHOSEN_CACHEPATH_REQUEST.\n" );
+       client_send_msg(fd, &message);
+       client_recv_msg_static(fd, &message, LDCS_READ_BLOCK);
```
mplegendre: We need to return out of this loop on network errors.

src/client/client_comlib/client_api.c
```diff
@@ -36,7 +37,56 @@ static struct lock_t comm_lock;
...
+       break;
+
+   }while( 1 );
+
+   if (message.header.type != LDCS_MSG_CHOSEN_CACHEPATH || message.header.len > MAX_PATH_LEN) {
+      err_printf("Got unexpected message of type %d\n", (int) message.header.type);
+      return -1;
+   }
+   char *local_crc = strdup( buffer );
+   char *local_cpc = strdup( &buffer[ strlen(local_crc) + 1 ] );
```
mplegendre: spindle_strdup instead of strdup on the above two lines

src/cobo/cobo.c
```diff
@@ -513,7 +513,7 @@ static int cobo_connect_hostname(char* hostname, int rank)
                       break;
                    case HSHAKE_INTERNAL_ERROR:
                       err_printf("Internal error doing handshake: %s", spindle_handshake_last_error_str());
-                      exit(-1);
+                      return -1;
```
mplegendre: Revert this back to exit(-1). 

The handshake is security checks. If things go wrong here, we just want to exit out and stop. We don't know if an attacker might be able to trigger the HSHAKE_INTERNAL_ERROR, and if they do we're better off just stopping the process.

src/cobo/cobo.c
```diff
@@ -837,7 +837,7 @@ static int cobo_open_tree()
               break;
            case HSHAKE_INTERNAL_ERROR:
               err_printf("Internal error doing handshake: %s", spindle_handshake_last_error_str());
-              exit(-1);
+              return -1;
```
mplegendre, line 840: Revert back to exit(-1)

src/server/auditserver/ldcs_audit_server_handlers.c
```diff
@@ -2950,6 +2960,93 @@ static int handle_client_pickone_msg(ldcs_process_data_t *procdata, int nc, ldcs
...
+/**
+ * Handle LDCS_MSG_REQUEST_CACHEPATH_CONSENSUS to determine which cachepaths are
+ * available across all of the servers.
+ */
+static int cachepath_consensus_reached;
+static int handle_cachepath_consensus(ldcs_process_data_t *procdata, ldcs_message_t *msg){
+
+    int num_children = ldcs_audit_server_md_get_num_children(procdata);
+
+    debug_printf( "Processing REQUEST_CACHEPATH_CONSENSUS.\n" );
```
mplegendre, line 2972: These debug_printfs are all at level1, but should be level 2-3.
level 1 should just be a single statement for the high-level operations. One or two "running consensus on ..." or "consensus reached on ..." would be appropriate.
level 2 should be the major operations within the high-level opertions. Many of these printfs should be level 2.
level 3 should be for the gross details.

src/utils/parseloc.c
```diff
@@ -168,38 +163,64 @@ char *parse_location_noerr(char *loc, number_t number)
...
    origpath = strdup(path);
-   for (;;) {
-      if (stat(origpath, &buf) != -1)
-         break;
-      if (cur_slash)
-         *cur_slash = '/';
+   errno=0;
+   while( stat( origpath, buf ) == -1 ){
+      local_errno = errno;
+      debug_printf("Failed to stat '%s' (%s).\n", origpath, strerror(local_errno));
```
mplegendre: Should probably be a debug level 3 printf

mplegendre (changes requested):

src/client/client_comlib/client_api.c
mplegendre: Also, the 10 second total may not be enough on large systems. Suggest significantly higher timeout.

src/client/beboot/spindle_bootstrap.c
rountree: Fixed.  Required changing `setup_environment()` to return `int` instead of `void`~~, and checking that return value in `main()`~~.

The return value is ignored, but that's a fix for another PR.  No longer checking for a NULL string; the check should occur in `send_cachepath_query()`.  If we can't trust the string that was handed to us, it that function needs to do a lot more checking (does the directory exist? is it writeable? etc.).

src/client/client/client.c
rountree: Same as above.

src/client/client/client.c
rountree: Assert removed.  Test not needed.

src/client/client/intercept_readlink.c
rountree: Removed assert.  No test needed.

src/client/client/should_intercept.c
rountree: Removed assert.  No test needed.

src/client/client_comlib/client_api.c
rountree: Agreed.  This was debugging code added to chase down what was ultimately a flux issue.  No need to keep it, but yes, remove it in a different PR.  
#187

src/client/client_comlib/client_api.c
rountree: Not sure if you want this retry/timeout ultimately removed, but I've changed it to an 0.1 second delay with 1000 retries.

src/client/client_comlib/client_api.c
rountree: Fixed.

src/client/client_comlib/client_api.c
rountree: Fixed.

src/server/auditserver/ldcs_audit_server_handlers.c
rountree: Yes.  What this is telling me is that I need a separate branch for debugging.

2 deleted, 7 changed to level 3, 4 changed to level 2, 1 retained at level 1.

src/utils/parseloc.c
rountree: Fixed for both in that function.

src/cobo/cobo.c
rountree: Fixed.

src/cobo/cobo.c
rountree: Fixed.

src/client/client_comlib/client_api.c
mplegendre: Maybe I'm confused. I thought this code was part of the race condition prevention for a client connecting before the consensus protocol finishes. Is the client not delaying for the consensus protocol to finish?

mplegendre (approved):

PR #194, opened 2026-06-22
Move cachpath consensus to be calculated at init time

### LINKED ISSUE #187: Change `send_cachepath_query()` to use state machine instead of retry.
[src/client/client_comlib/client_api.c](https://github.com/llnl/Spindle/pull/182/files/b5970ec4d2e97c3742f50e2c143adecaa689f98f#diff-a0629be54afba71122077845eefb066a40860c6b2f97a93d4ea0c4f064a3ad44)

`send_cachepath_query()` has a bunch of timeout and retry code that was added when debugging another problem.  Remove this code after the commpath PR gets in.

mplegendre:
This fixes #187 

Cachepath consensus was being triggered by a message from the FE. This was the likely cause of race conditions as the consensus algorithm collided with both the client connections and other communications happening. 

This comment removes the trigger for consensus being a message initiated by the FE. Instead consensus is triggered during server initialization, before clients are allowed to connect. Because of this we were able to remove the timeout/retry code from clients that triggers if they connect before consensus is reached.


rountree:
Code looks good, works 3x on rzadams.  Will go ahead and merge.

PR #196, opened 2026-07-06
[multiple-commpaths] Allows colon-separated commpath list

rountree:
Per-server selection, first path in `--commpaths` that works for that server gets picked.  No consensus process, as the commpath is not visible outside of the server and its particular clients.


mplegendre:
@rountree - Are you working on fixing the CI failures?

rountree:
@mplegendre Yes.  What's queued up for this coming Friday is:

1. A PR adding workflows for granular resource manager testing and log recovery via artifacts.
2. A FAQ on how to launch GitHub workflows, monitor progress, and download artifacts all via the command line.
3. A PR with most of the low-hanging-fruit configure file changes.
4. A patched multiple-commpaths PR that passes repeated invocations of the existing `ci.yml` workflow.

rountree:
Debugging notes:

Running `bd856fa` (tip of `multiple-commpaths` at the moment)
3/13 `ci.yml` failures
a) Flux, `Running: ./run_driver --dependency --push`, `Error: The operation was canceled.` 17m 27s, `31038941417`
b) slurm-rshlaunch, `SOME TESTS FAILED` (no failures logged), 2m 9s, `31038944077`
c) slurm-srun, `SOME TESTS FAILED` (no failures logged), 1m 18s, `31038929449`

Running `c0170276` (tip of `devel`)
3/31 `ci.yml`
a) Flux, `Running: ./run_driver --dependency --push`, `Error: The operation was canceled.`, 17m 7s, `31043385441`
b) ditto
c) ditto

Running `bd856fa` (multiple-commpaths)
0/31 `ci-slum-srun-artifacts` failures
0/31 `ci-slurm-artifacts` failures

----------

Running `bd856fa` (multiple-commpaths)
4/13 `ci.yml` failures, 3 Flux, 1 `Run spindle-slurm-ubuntu testsuite`

----------
Running `c0445ae` (testing-multiple-commpaths)
Run `31106514887` job `92632975191` "Slurm srun, Ubuntu"

At this point GitHub actions became unusably slow.

The good news is that the podman port can replicate the problem on rzadams with a 4% failure rate.

.github/workflows/ci-slurm-plugin-artifacts.yml
```diff
@@ -0,0 +1,123 @@
...
+        id: slurm-ubuntu-testsuite-sessions
+        run: |
+          docker exec slurm-plugin-head bash -c 'cd Spindle-build/testsuite && SPINDLE_DEBUG=3 salloc -n${workers} -N${workers} --spindle-session ./runTests ${workers}'
+
+      - name: Collect Spindle logs from containers
+        if: ${{ always() }}
+        continue-on-error: true
+        run: |
+          failed=0
+          for node in slurm-head slurm-node-1 slurm-node-2 slurm-node-3 slurm-node-4; do
```
nchaimov:
```suggestion
          for node in slurm-plugin-head slurm-plugin-node-1 slurm-plugin-node-2 slurm-plugin-node-3 slurm-plugin-node-4; do
```
I think this will need to use the container name as defined in the `docker-compose.yml`, which for the plugin variant includes `-plugin-` in the name (see for example `docker exec slurm-plugin-head` above)

src/slurm_plugin/plugin_utils.c
nchaimov:
```suggestion
   char *realized_dir = NULL, *socket_path = NULL;
```
With the change below, if `getFirstValidPath` returns -1, we `goto done` without ever assigning to `realized_dir`; before these changes, we would always assign to `realized_dir`. Since the `done` block checks the value of `realized_dir`, we need to initialize it to `NULL`.

src/fe/startup/parse_launcher.cc
```diff
@@ -292,6 +292,7 @@ void ModifyArgv::modifyCmdLine()
    snprintf(options_str, 32, "%lu", (unsigned long) params->opts);
    string options(options_str);
    
+   getFirstValidPath( params->commpaths, &( params->commpath ), params->number );
```
nchaimov: Need to handle the case where `getFirstValidPath` returns -1

configure.common.ac
```diff
@@ -29,18 +29,18 @@ AC_ARG_WITH(cachepath,
             [[],[]],
             [AC_MSG_ERROR(use --with-cachepaths=DIRS (plural) instead of --with-cachepath=DIR to specify one or more cache paths)],
             [])
-AC_ARG_WITH(commpath,
-            [AS_HELP_STRING([--with-compath=DIR],[Back-end directory for communication and housekeeping])],
-            [COMMPATH=${withval}],
-            [COMMPATH=$DEFAULT_LOC])
+AC_ARG_WITH(commpaths,
+            [AS_HELP_STRING([--with-compaths=DIR],[Colon-separated list of candidate paths for back-end communication and housekeeping])],
```
nchaimov:
```suggestion
            [AS_HELP_STRING([--with-commpaths=DIR],[Colon-separated list of candidate paths for back-end communication and housekeeping])],
```
The name of the argument is `commpaths` but the help string is `--with-compaths`, with one m instead of two.

src/flux/sessionmgr.c
```diff
@@ -106,27 +106,14 @@ char **strip_start_from_argv(int argc, char **argv)
    return new_argv;
 }
 
-extern char *parse_location(char *loc, int number);
-extern int spindle_mkdir(char *orig_path);
-
-#if !defined(COMMPATH)
-#error COMMPATH must be defined in config.h
-#endif
+#define SESSIONPATH "/tmp/spindle/session"
```
nchaimov: I'm not sure we should hard-code a path for the session directory

src/flux/sessionmgr.c
```diff
@@ -106,27 +106,14 @@ char **strip_start_from_argv(int argc, char **argv)
...
 {
-   int result;
-   char *dir;
-   dir = parse_location((char *) (COMMPATH "/spindle_session"), 0);
-   if (!dir) {
-      spindle_debug_printf(1, "ERROR: Could not parse directory for spindle session location from %s/spindle_session\n", COMMPATH);
-      return NULL;
-   }
-      
-   result = spindle_mkdir(dir);
```
nchaimov, line 125: The old code called `spindle_mkdir` to ensure that the session directory exists, but the new code doesn't. Do we still need the `spindle_mkdir`?

configure.common.ac
```diff
@@ -29,18 +29,18 @@ AC_ARG_WITH(cachepath,
             [[],[]],
             [AC_MSG_ERROR(use --with-cachepaths=DIRS (plural) instead of --with-cachepath=DIR to specify one or more cache paths)],
             [])
-AC_ARG_WITH(commpath,
-            [AS_HELP_STRING([--with-compath=DIR],[Back-end directory for communication and housekeeping])],
-            [COMMPATH=${withval}],
-            [COMMPATH=$DEFAULT_LOC])
+AC_ARG_WITH(commpaths,
```
nchaimov, line 32: We might want to still accept the old `commpath` (singular) form so that existing configure arguments continue to work

src/client/client_comlib/client_api.c
nchaimov: The above line increased the size of the buffer, but this line still null-terminates at the original size; is that correct?

rountree: Looks like Matt doubled the buffer length.  Git blame says:
```
68bacdf8 src/client/client_comlib/client_api.c (Matthew LeGendre 2026-06-22 11:17:35 -0700  45)    char buffer[2*(MAX_PATH_LEN+1)];
af4cd951 src/client/client_comlib/client_api.c (Barry            2025-10-03 09:52:29 -0700  46)    buffer[MAX_PATH_LEN] = '\0';
```
The buffer is only used for the reply message (`LDCS_MSG_CHOSEN_CACHEPATH`) which copies two directory names into the buffer for a total length of `strlen(procdata->cachepath) + 1 + strlen(procdata->parsed_cachepath) + 1`.  There's no check on the sending side if that value is `<= 2*(MAX_PATH_LEN+1)`, although there may be an implied check upstream.

The line adding the null terminator is not doing anything useful.  Adding one to the end of `buffer` might prevent a segfault if the incoming strings were malformed but otherwise fit into the buffer.

On my own projects I use `calloc()` instead of `malloc()` unless there's a good reason not to.  Your preference?

rountree: Fixed per discussion:  buffer truncation now occurs at the correct address and is applied after the buffer is filled.

src/slurm_plugin/plugin_utils.c
rountree: Nice catch.

src/fe/startup/parse_launcher.cc
rountree: Both `getFirstValidPath` and `getApplicationArgsFE` can fail with return values of `-1`.  Now checking both.  Modified the enclosing function to return `int` instead of `void`, and updated its call sites to check the return value.

configure.common.ac
rountree: Both singular and plural forms now accepted for both commpath[s] and cachepath[s], and for configure, build, and runtime options.

src/flux/sessionmgr.c
rountree: This one got complicated quickly.  The naive implementation (picking a path from `--commpaths`) requires the same `--commpaths` parameter used for the session initialization job be present in every job to be run under that session, and also requires that the state of the candidate paths doesn't change of the course of the session (e.g., if the mount described by the first path in the list was initially down but came back online later, jobs fail).

Possible fix is to have a dedicated `--sessionpaths` parameter that's evaluated only at session start, and the `SESSION_ID` returned by the initializing job embeds the path to use.  If no `--sessionpaths` are working, then fall back to `commpaths`, and eventually `$TMPDIR`.

Will take this up in a separate PR.

For now, change the hard-coded path to be based on `$TMPDIR`.

rountree:
@nchaimov Ready for review.

nchaimov (approved):
It looks like this addresses all of my comments from the previous review

PR #197, opened 2026-08-01
Crash handler logging, documentation

nchaimov:
This PR extends the crash handler:

- The alternate stack to handle crashes due to stack overflow is now optional and off by default. It is enabled with `--crash-altstack`. In the event of a stack overflow without an alternate stack, the kernel will restore the default signal handler automatically and core dumps will not be deduplicated.
- The crash handler now optionally produces a log showing which ranks crashed at which sites. This is enabled with `--crash-log=[path]`. If the optional path is not provided, the log will be written to `spindle-crash-log.$NUMBER` in the working directory; if it is provided, it will be written at the provided path.
- To enable use with the Slurm plugin, this PR also adds the ability to specify Spindle arguments when starting a session through `salloc` and `sbatch`; this functionality was previously missing.
- The Flux plugin now handles the new arguments.
- A new documentation file `doc/spindle_crash_handler_README.md` describes the use of the crash handler.
- The manpage is updated with the new arguments.
- The launcher documentation `doc/spindle_launch_README.md` is updated with the new options.
- The crash handler tests now include session mode tests and validation of crash log files.

The crash handler logging is implemented as follows: previously, as soon as a server recognized a crash site it had previously seen, it immediately sent a not-selected reply and did not pass the message further up the tree. Now, if crash logging is enabled, the not-selected reply is still sent immediately, but an additional message is passed up the tree to report an additional crashing rank. Crash reports also now include a display rank representing the MPI or scheduler-provided rank as determined from environment variables. The root server maintains a table of crash sites and associated display ranks. When the root server exits, it produces a log file having the format:

`exe: /home/me/my_app
site: libsolver.so.1+0x2f10
exemplar: 4
count: 13
ranks: 4-15,18

exe: /home/me/my_app
site: abort:my_app: solver.c:88: solve: Assertion `n > 0' failed.
exemplar: 0
count: 4
ranks: 0-3
`

Each entry lists the executable that crashed, the library+offset or assert message associated with the failure, the rank of the exemplar core dump that was selected to be output, the total number of ranks that crashed at that crash site, and a comma-separated list of ranges of ranks that crashed at that site.

Because the log is written when the server exits, when session mode is used, a single log is produced for the whole session, with coredumps deduplicated across the whole session. To handle multiple executables being run within a session, the executable is now always included in the crash key so that crashes at the same site in a given library called from two different executables will not be deduplicated. 

Two new tests are added to verify behavior in session mode: one test in which the same executable crashes in the same site in two different jobs within a session, and another test in which two different executables crash at the same site in a library. The crash tests now verify that the expected crash sites and ranks are recorded in the log file and match the crash site in the core dump.


nchaimov:
New changes to this PR:

- The bug where we were recording a crash in the log for `kill(getpid(), SIGSEGV)` but not getting a coredump, and seeing a delay on exit, is fixed. I had assumed that returning from the signal handler would always cause the signal to recur, but this is only true for signals triggered by an actual fault. User-sent signals are not retried upon return from the handler. This is fixed by checking the cause of the signal and explicitly re-raising it if it wasn't a kernel-triggered signal. With this fix, we are no longer missing the coredump and there is no delay on exit.
- The crash log is now in CSV format, with one row per crashing rank. 
- The crash log now includes the predicted path to the core dump that will be produced for the exemplar for the rank's crash site.

nchaimov:
New changes:

- Handle case where the application-set disposition is `SIG_IGN`. For a user-sent signal, such as from `kill()`, these are now ignored. For kernel-sent `SIGSEGV`, these are treated as `SIG_DFL`, matching the kernel behavior.

- Fixes crash handler in presence of fork. Previously, the fork child would share the same crash handler state as the parent. Now, when the first Spindle-handled call occurs after a fork, the crash handler is re-initialized and locks are released. If a crash occurs between the fork and the first Spindle-handled call, the crash handler will fall back to not handling the crash. The crash log file now contains a pid column to distinguish forked children that would otherwise be attributed solely by the rank shared with the parent. New fork test modes were added.

- If a crash occurs on the thread holding the `comm_lock`, it is not safe to communicate in the crash handler, as this thread could have been in the middle of a communication. If the crashed thread holds the `comm_lock`, we now fall back to not handling the crash.

- The hostname and a timestamp are now included in the crash log.

- If the chosen crash log path already exists, we now append to it instead of overwriting it. If the file was created, a CSV header is written first; if the file already existed, the header is skipped.

- A bug was fixed where the crash handler would open libraries by original path instead of relocated path when determining file offset. The crash handler used the path from `l_name`, which was the original path. The relocated path is now read from `/proc/self/maps` when opening the file, but the value from `l_name` is still used as the crash site key and in logging, so that the original path is logged.

- The executable path used as the crash site key for deduplication, and written in the log, is now the original path rather than Spindle's relocated path.

nchaimov:
I have run the E4S test suite with this PR. The same failures seen on devel also occur on this PR (see #215), but no additional failures are introduced.

PR #203, opened 2026-08-04
ci artifacts

rountree:
Splits the ci.yml workflow into individual jobs, all of which run
    with SPINDLE_DEBUG=3.  Logs created regardless for all runs (for
    now) and copied into downloadable artifacts.

    The crash testing code was not copied over.

    Code initially generated by Claude Code + Opus, then manually
    debugged.

    There are any number of improvements possible.  Some require
    reworking runTests, some (like repeated runs) can be managed with
    the gh command-line utility, and some just require additional
    code complexity (default logging on failure, optional preserving
    logs on all runs).  This PR is intended to be a solid baseline
    to start from.


rountree:
Not quite fully baked.

PR #213, opened 2026-09-16
Fix issues around glibc mis-calculating IE TLS sizes when spindle is running

mplegendre:
GLIBC has a bug where any LD_AUDITer disables the normal calculation for an application's IE TLS (also called static TLS in this commit). Applications that need more than the default amount of IE TLS will fail with a message about insufficient TLS under spindle. Based on conversations with GLIBC folks, this may not be fixed for a while (or ever).

To work around this, this commit manually calculates the amount of IE TLS needed before each exec() operation, then if more than the default it uses the GLIBC_TUNABLES environment variable to set the IE TLS amount. Spindle does this scalably by:
- In the client, before an exec, collect the executable, LD_LIBRARY_PATH, LD_PRELOAD, and component of PATH for an exec, the set of which is used to uniquely identify an exec and its runtime environment.
- Send that info to the server, which does a standard spindle reduction/broadcast operation to perform I/O on a designated server and share results.
- Using the provided environment, run an equivalent of 'ldd app' (but using direct calls to ld-linux.so) to get the list of libraries loaded at startup.
- Iterate over the libraries, executable, and spindle runtime to read their TLS requirements from the binaries and calculate the total IE TLS needed. 
- Broadcast results back down through servers and clients.

This PR also adds several libraries (libbigtls.*) to the testsuite's set of library loads. These libraries have enough TLS to overrun the default IE TLS provided by glibc and require the above algorithm to kick in.


scripts/static_tls_test/run_test.sh
```diff
@@ -0,0 +1,18 @@
...
+
+mkdir -p $BINDIR
+
+if [ ! -f "$BINDIR/statictls" ]; then
+$CC $CFLAGS -o $BINDIR/libstatictls.so -shared -fPIC $SRCDIR/statictls.c
+$CC $CFLAGS -o $BINDIR/statictls -L$BINDIR -lstatictls -lpthread -Wl,-rpath,$BINDIR $SRCDIR/app.c
+$CC $CFLAGS -o $BINDIR/libminaudit.so -fPIC -shared $SRCDIR/minaudit.c
+fi
+LD_AUDIT=$BINDIR/libminaudit.so $BINDIR/statictls >& /dev/null
+RESULT=$?
```
nchaimov: This line is producing a syntax error when run by configure:
```
/home/spindleuser/Spindle/scripts/static_tls_test/run_test.sh: 16: Syntax error: Bad fd number
```
See, for example, the build of the serial image: https://github.com/llnl/Spindle/actions/runs/35164411510/job/105022155537#step:4:1050

I think the problem is the `>& /dev/null` in the previous line. The script specifies `#!/bin/sh` but that form of redirect is a bash extension.

src/fe/startup/config_mgr.cc
```diff
@@ -158,6 +158,12 @@ using namespace std;
 #define PYTHON_PREFIX_DEFAULT ""
 #endif
 
+#if defined(STATIC_TLS_ALLOC_BUG)
```
nchaimov: `STATIC_TLS_ALLOC_BUG` is always defined to something, as the configure script always does
```bash
16902 cat >>confdefs.h <<_ACEOF
16903 #define STATIC_TLS_ALLOC_BUG $STATIC_TLS_BUG
16904 _ACEOF
```
where `$STATIC_TLS_BUG` is either 0 or 1, so `--calc-tls` will be true by default regardless of the result of the test in configure.

nchaimov:
I've run the E4S testsuite against this PR and found that there are five tests that are failing with this PR that pass without it:

Fail:
- `hpctoolkit`
- `libpressio`
- `openfoam`

Timeout:
- `tau`: times out at 600s, in devel passes in 7s
- `xyce`: times out at 600s, in devel passes in 136s

Working on collecting more information on these failures

